PR #2, opened 2018-02-22
Boilerplate Rails

afred:
This is the result of running `rails new`.


afred:
@WGBH-MLA/developers please review and merge when you get a chance. Lots of files, but it's all boilerplate code. Given that the Hyrax install is also touching a lot of files, I wanted to break the initial setup into 2 steps, so the latter cold be reviewed with much less noise.

PR #23, opened 2018-03-12
AMS2-58: Sets up Amazon SES for Production

jasoncorum:
Closes AMS2-58.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -1,4 +1,10 @@
 Rails.application.configure do
+
+  # Method for using environment variables for Booleans
+  def truthy_env_var?(val)
+    ['yes', 'true', '1'].include? val.downcase.strip
+  end
+
   # Settings specified here will take precedence over those in config/application.rb.
 
   # Code is not reloaded between requests.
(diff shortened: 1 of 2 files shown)


config/environments/production.rb
```diff
@@ -68,6 +68,22 @@
...
   config.action_mailer.default_url_options = { :host => ENV['PRODUCTION_HOST'] }
+  # Settings for emails through Amazon SES
+  # Replace with environment variable
+
+  config.action_mailer.delivery_method = :smtp
+  config.action_mailer.perform_deliveries = true
+  config.action_mailer.raise_delivery_errors = false
+  config.action_mailer.default :charset => "utf-8"
+
+  config.action_mailer.smtp_settings = {
```
adeelahmad: Can we put all SMTP params in the environment?

jasoncorum: As in define them all as environment variables? Sure, just more variables we'll have to figure out how to define when setting up the environment which I feel like can be a PITA.

adeelahmad: I agree this would be an issue but we can put this into a How-to/ReadME.
The problem is not everyone is going to use Amazon SES or port 587 and can be an issue for those people if we hard code it.

jasoncorum: Good point!

jasoncorum: adeel - how do you feel about strictly defining `enable_starttls_auto: true`? i think if we want to set this as an envar we'd need to come up with a clever way to cast a string value to a ruby True or False. there's a wannabe_bool gem, but it's not available at this point. thoughts?

PR #28, opened 2018-03-15
AMS2-21 Adding user role as manger to admin set

adeelahmad:


afred:
Hmm.. Net Timeout error on that last build. Restarted it.

PR #36, opened 2018-03-21
fixing deployment issues

adeelahmad:
fixes AMS2-104
fixes deployment html page

### DIFF
--- a/bin/deploy/install.sh
+++ b/bin/deploy/install.sh
@@ -1,13 +1,20 @@
 #!/bin/bash
+echo "Running Deployment for ID: $DEPLOYMENT_ID"
+export HOME=/var/www/ams
+cd $HOME
 source /etc/profile
-cd /var/www/ams
+echo "ruby version:`ruby -v`"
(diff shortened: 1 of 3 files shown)


bin/deploy/install.sh
```diff
@@ -1,13 +1,20 @@
...
-chown -R ec2-user:ec2-user /var/www/ams
+sudo chown -R ec2-user:ec2-user /var/www/ams
 if [ -z $SECRET_KEY_BASE ]; then
-echo "export SECRET_KEY_BASE=`rails secret`" >> /etc/profile
+sudo echo "export SECRET_KEY_BASE=`rails secret`" >> /etc/profile
 source /etc/profile
 fi
-bin/deploy/create_deployment_details_page.rb
-service httpd restart
+DEPLOYMENT_ID=$DEPLOYMENT_ID ruby bin/deploy/create_deployment_details_page.rb
```
afred: @adeelahmad why are we setting the ENV var to itself here. Line 2 makes it look like it's already set, right?

PR #40, opened 2018-04-02
AMS2-11 add new work asset

adeelahmad:
Adds

- adds new work type Asset
- adds custom presenter for single asset view
- adds new multiple date input for simple form
- override default form edit fields for date view
- change hyrax config to allow creating work without
  - files
  - deposit agrement
- Modify blacklight catalog controller for
  - Search Result
  - Sorting
  - Single Display


spec/features/create_work_spec.rb
```diff
@@ -62,7 +62,7 @@
       find('body').click
       choose('work_visibility_open')
       expect(page).to have_content('Please note, making something visible to the world (i.e. marking this as Public) may be viewed as publishing which could impact your ability to')
-      check('agreement')
+
```
afred: This seems odd. Why don't we have to agree to the TOS here?

adeelahmad: Just removed agreement checkout (within hyrax config) for a faster asset upload for now.

adeelahmad:
@afred I have cleaned up the code. Can you review now?

PR #42, opened 2018-04-06
AMS2-146 Adding genre vocabulary

adeelahmad:


spec/features/create_asset_with_genre_vocabulary_spec.rb
```diff
@@ -0,0 +1,76 @@
+require 'rails_helper'
+include Warden::Test::Helpers
+
+RSpec.feature 'Create Asset with Asset Type', js: true do
+  context 'Create adminset, create asset' do
+    let(:admin_user) { create :admin_user }
```
afred: `admin_user` doesn't appear to be used anywhere else.

afred:
The only small change needn't block the merge.

Also, I'm not sure we need to maintain 2 specs for creating an asset, one with genre, and the other with everything except genre. Could they be merged into a single spec?

PR #44, opened 2018-04-09
Merge from develop

afred:


adeelahmad:
@afred Shouldn't we be merging into master?

afred:
Ah.. i right. Stand by.

PR #46, opened 2018-04-10
AMS2-17 Adds Physical Instantiation Work Type

adeelahmad:


spec/features/create_physical_instantiation_spec.rb
```diff
@@ -0,0 +1,105 @@
+require 'rails_helper'
+include Warden::Test::Helpers
+
+RSpec.feature 'Create and Validate Series', js: true do
+  context 'Create adminset, create series' do
```
afred: Need to replace "Series" verbiage with "Physical Instantiation"

config/locales/physical_instantiation.de.yml
```diff
@@ -0,0 +1,8 @@
+de:
+  hyrax:
+    icons:
+      physical_instantiation:     'fa fa-file-text-o'
```
afred: Do we know what `fa fa-file-text-o` is supposed to be? Is it an HTML ID or class value? And if so, why is it in a localization file?

adeelahmad: @afred i believe `fa fa-file-text-o` is css class for file icon.

PR #51, opened 2018-04-18
AMS2-17 adds physical instantiation

adeelahmad:


afred:
@adeelahmad Had to resolve conflicts, but once this is green, we should be good to go.

PR #55, opened 2018-04-20
DO NOT MERGE - JUST FOR REFERENCE - Adds association between Asset and Series

afred:
* Association is a 'has and belongs to many'.
* Adds association to models.
* Modifies Series form to make adding Assets clearer, changes labels, etc.
* Adds an authority for returning Assets that can be added to a Series.
* Creates search builder for returning Assets that can be related to Series.
* Adds custom actor for adding Assets to Series.
* Modifies default actor stack to use custom actor.

Also,

* Fixes Series factory to include description (a required field).


afred:
Closing because this is unlikely to get merged.

PR #58, opened 2018-05-02
AMS2-140 import digital instantiation essence track from pbcore xml

adeelahmad:


app/forms/hyrax/digital_instantiation_form.rb
```diff
@@ -3,9 +3,26 @@
 module Hyrax
   class DigitalInstantiationForm < Hyrax::Forms::WorkForm
     self.model_class = ::DigitalInstantiation
+    self.terms += [:digital_instantiation_pbcore_xml]
```
jasoncorum: @adeelahmad - i'm curious, why is this broken out in to its own `self.terms +=`?

adeelahmad: @jasoncorum nothing really, got in there for testing, cleaned it up. Can you review again?

PR #62, opened 2018-05-08
Adds typed title field

afred:
* Adds distinct properties for various title types to the Asset model.
* Creates a custom input that combines a text inputs with a selet box for the
  type.
* Modifies AssetForm object to split distinct properties for each title type
  into a structure suitable for displaying as a combination of select box (for
  the title type) and a text box (for the title).
* Adds logic to AssetActor to modify the environment on #create and #update. The
  modifications combine the title and title type fields from the submitted
  form data into distinct properties for each title type on the model before
  saving.
* Modifies form partials for edit fields to allow for multiple title/title_type
  paris, and ensure they are pre-populated.
* Modifies AssetPresenter, relevant view partials to display all the different
  title types.
* Adds a controlled vocabulary for title and description types.
* Adds helper methods for dealing with Asset create/update forms in Capybara.

Also,

* MultipleTextSelectComboInput#collection overrides MultiValueInput#collection
  (from hydra_editor). This may be a useful patch to hydra_editor.

### DIFF
--- a/app/actors/hyrax/actors/asset_actor.rb
+++ b/app/actors/hyrax/actors/asset_actor.rb
@@ -3,6 +3,52 @@
 module Hyrax
   module Actors
     class AssetActor < Hyrax::Actors::BaseActor
+      def create(env)
+        add_title_types(env)
+        add_description_types(env)
+        super
+      end
+
+      def update(env)
+        add_title_types(env)
+        add_description_types(env)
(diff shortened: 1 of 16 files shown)


afred:
I'm getting some errors locally on this locally that may be due to my not-so-fast computer, because when I try to debug it, they pass.

PR #66, opened 2018-05-09
Adds 'program' type for titles and descriptions

afred:


app/models/solr_document.rb
```diff
@@ -229,8 +229,12 @@ def alternative_modes
     self[Solrizer.solr_name('alternative_modes')]
   end
 
+  def program_title
+    self[Solrizer.solr_name('program_title')]
+  end
+
   def episode_title
-    self[Solrizer.solr_name('episode_title')]
+    self[Solrizer.solr_name('program_title')]
```
adeelahmad: @afred don't we need `episode_title` here?

jasoncorum: @adeelahmad - good catch!

PR #69, opened 2018-05-14
240 typed field css

afred:
wait until #68


afred:
style changes have the following effect

image

image

PR #90, opened 2018-06-11
Adds initializer for locale files

jasoncorum:
This will require that the default CollectionType rake task will need to be re-run on edge and staging instances.

`rails hyrax:default_collection_types:create`


adeelahmad:
@jasoncorum We might have to clean-out dev and staging server db to create the collections again.

PR #92, opened 2018-06-19
218 Assets Only Creation from Create Work

jasoncorum:
**Need to merge after PR for 181**

A few notes on this PR:

1. It removes specs that simply created objects other than an Asset from the "Add new work" button in favor of create other objects through the Asset.
2. It overwrites select_type_list_presenter to limit model creation.
3. This removes create_contribution_spec.rb as it is outdate since it creates the contribution from the general "Add new work button." We will rewrite this test in the next sprint to complete AMS2-181.


app/presenters/hyrax/select_type_list_presenter.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    class_attribute :row_presenter
+    self.row_presenter = SelectTypePresenter
+
+    # @return [Boolean] are there many differnt types to choose?
+    def many?
+      authorized_models.size > 1
+    end
+
```
afred: Can we add these comments?
```
# Overwritten from Hyrax's app/presenters/hyrax/select_type_presenter.rb.
# We need a different implementation of #authorized_models to only return our Asset model.
```

PR #100, opened 2018-07-06
Sprint 9

adeelahmad:


spec/models/essence_track_spec.rb
```diff
@@ -1,9 +1,160 @@
 # Generated via
-#  `rails generate hyrax:work EssenceTrack`
+#  `rails generate hyrax:work Asset`
```
afred: remove inaccurate comment

spec/models/asset_spec.rb
```diff
@@ -3,7 +3,295 @@
...
+  context "clip_description" do
+    let(:asset) { FactoryBot.build(:asset) }
+    it "has clip_description" do
+      asset.clip_description = ["Test clip_description"]
+      expect(asset.resource.dump(:ttl)).to match(/pbcore.org#hasClipDescription/)
+      expect(asset.clip_description.include?("Test clip_description")).to be true
+    end
+  end
+
+
```
afred: The factories used in the model specs can be defined once at the top-level `describe`, e.g. `let(:asset) { FactoryBot.build(:asset) }`, rather than per context.

bin/deploy/install.sh
```diff
@@ -1,7 +1,9 @@
 #!/bin/bash
 echo "Running Deployment for ID: $DEPLOYMENT_ID"
+sudo yum -y update
 export HOME=/var/www/ams
 cd $HOME
+rvm gemset use ams --create
```
afred: just curious, what does using a gemset here do for us?

PR #106, opened 2018-07-18
AMS2-430 restrict admin data to group

adeelahmad:


afred:
Merging despite a known bug in an integration test, the web driver is unable to select from a drop down populated by an AJAX request. It works locally, just not on Travis.

PR #107, opened 2018-07-18
Adds custom icons

afred:
Custom icons are for Asset, DigitalInstantiation, PhysicalInstantiation, and EssenceTrack.

However, these do not cover the default thumbnails, which is probably what we'll want eventually.

for now, closes AMS2-193


afred:
Here's what they look like....

**Asset**
image


**Digital Instantiation**
image


**Physical instantiation**
image


**Essence Track:**
image

PR #109, opened 2018-07-23
Changes 'episode number' field to be a title type.

afred:
Also, runs migrations if needed in AMS.reset_data!


afred:
@adeelahmad this is currently failing for me locally, but it seems to be a web driver issue, so trying it on Travis to see what happens.

afred:
Hmm.. ok, looks like it's not just a web driver issue.

Gemfile
```diff
@@ -49,6 +49,8 @@ group :development, :test do
   gem 'bixby' # bixby == the hydra community's rubocop rules
   gem 'selenium-webdriver'
   gem 'chromedriver-helper'
+  # Alternative web driver to selenium; requires phantomjs
+  gem 'poltergeist'
```
adeelahmad: @afred can we get rid of this, if we are not using this anymore ?

afred: yep, sorry, stand by

PR #118, opened 2018-08-08
AMS2-479 Adding Producing Organization to asset

adeelahmad:


config/authorities/contributor_role.yml
```diff
@@ -3,7 +3,7 @@ terms:
     term: Actor
   - id: Artist
     term: Artist
-  - id: Artistic Director
+  - id: Artistic Directorx
```
afred: @adeelahmad typo?

PR #124, opened 2018-08-15
Different approach to Asset-only search results

afred:
Adds a custom SearchBuilder rather than modifying the base SearchBuilder class.

Fixes AMS2-401.


adeelahmad:
@afred can we remove `Type` facet from Catalog Controller?

afred:
@adeelahmad Yes, stand by.

adeelahmad:
@afred you can merge once all tests pass.

PR #132, opened 2018-08-23
Adds date range search filter

afred:
Also,
  * Removes redundant solr/conf.


afred:
image


image


image


image

PR #165, opened 2018-10-23
Add batch ingest dependency

afred:


afred:
oops.. this was branched off of `master` rather than `develop`. Closing it, and will resubmit a PR based off of `develop`

PR #173, opened 2018-11-15
555 Adds sony_ci_api gem

jasoncorum:
Note: The sony_ci_api gem requires config/ci.yml with our Sony Ci credentials to be passed in to it. We will need to add that file to the servers for this to work on edge and staging.


app/controllers/media_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def show
+    ci = SonyCiBasic.new(credentials_path: Rails.root + 'config/ci.yml')
+    redirect_to ci.download(@solr_document['sonyci_id_tesim'][(params['part'] || 0).to_i])
+  end
+
+  private
+
+  def set_solr_document
+    @solr_document = ActiveFedora::Base.find(params['id']).to_solr
```
adeelahmad: @jasoncorum  do we need to make sure the user have access to the asset?

afred:
@adeelahmad @jasoncorum I added specs for the MediaController and for the "/media/[id]" route. I think we need a test for passing in the `&part=1` part of the URL, but I'm not entirely sure how it works.

The way the controller was written, it appears that if a media file has multiple parts, then each Sony Ci ID is contained within the `sonyci_id_tesim` field on the Solr Document. Is that correct?

afred:
@jasoncorum I can't add you as a reviewer, since you submited the PR. But could look this over and give a 👍 or 👎 anyway?

PR #176, opened 2018-11-21
87 index thumbnails

jasoncorum:
This creates an AssetThumbnailPathService that looks at an asset's digital_instantiations and applies the correct location for the source of that asset's thumbnail.


spec/indexers/hyrax/asset_indexer_spec.rb
```diff
@@ -24,11 +39,71 @@
...
-      work_type = solr_document.fetch('has_model_ssim').first.downcase
-      default_image = ActionController::Base.helpers.image_path(work_type+'.png')
-      expect(solr_document.fetch('thumbnail_path_ss')).to eq default_image
+    context "an asset that has digital_instations with AAPB defined as the organization" do
+      context "and 'Moving Image' defined as the media_type" do
+        it "has a S3 thumbnail if it has a sony_ci_id" do
+          asset.ordered_members << aapb_moving_image_digital_instantiation_work
+          asset.save
+          solr_doc = asset_solr_doc.generate_solr_document
+          default_image = ActionController::Base.helpers.image_path("http://americanarchive.org.s3.amazonaws.com/thumbnail/#{solr_doc[:id]}.jpg")
```
afred: @jasoncorum is the `#image_path` helper needed when we're dealing with a full URL?

PR #183, opened 2018-11-26
Uses shared specs from batch ingest

afred:
* Adds shared specs from hyrax-batch_ingest gem to test ZippedPBCoreReader class.

Also...

* DRY up spec for ZippedPBCoreReader a little.
* Fix FactoryBot deprecation warnings re: static attributes.
* Removes permisstion_templates.rb factory (nothing in there).


afred:
fyi, changing the name of the variables in the `let` statements it because the shared specs expect some variables to be set, i.e. `source_location` and `invalid_source_location`

PR #186, opened 2018-12-04
AMS2-564 export valid pbcore 2.1

adeelahmad:


adeelahmad:
@afred can you review this again.

PR #187, opened 2018-12-05
Add space to Asset Type id 

sroosa:
Some edit.s to controlled vocabularies, including avoiding quotation marks and adding spaces to ids, where we missed that before

### DIFF
--- a/config/authorities/format.yml
+++ b/config/authorities/format.yml
@@ -39,10 +39,14 @@ terms:
     term: DVCPRO
   - id: DVD
     term: DVD
-  - id: "Film: 16mm"
-    term: "Film: 16mm"
-  - id: "Film: 35mm"
-    term: "Film: 35mm"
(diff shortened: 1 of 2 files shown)


config/authorities/format.yml
```diff
@@ -39,10 +39,14 @@ terms:
...
-    term: "Film: 16mm"
-  - id: "Film: 35mm"
-    term: "Film: 35mm"
+  - id: >
+      Film: 16mm
+    term: >
+      Film: 16mm
+  - id: >
+      Film: 35mm
+    term: >
```
adeelahmad: @sroosa any reason we are using `>` in here?

sroosa: I didn't want the the quotes, but if you don't use the > then it doesn't like that the : in the term. If you prefer the other way, I can deal with it.

PR #188, opened 2018-12-12
Updates bundle

afred:
Needed the latest version of hyrax-batch_ingest.


afred:
@adeelahmad No, I amended the commit message but apparently forgot to put the changed db schema in as part of it. Just pushed again.

afred:
Ok.. that didn't do the trick either. Same `NoMethodError` for `repo_object_id` on a `BatchItem` instance in the shared spec for `BatchReader` objects. Running locally to look closer...

afred:
After doing `RAILS_ENV=test bin/rails db:environment:set db:reset db:migrate` on my local machine, this passes... so @adeelahmad you must be right about caching after all.

PR #189, opened 2018-12-13
Adds CSV export reports for Digital & Physical Instantiations

jasoncorum:
Closes 566 and 567.

### DIFF
--- a/app/models/ams/csv_export_extension.rb
+++ b/app/models/ams/csv_export_extension.rb
@@ -1,13 +1,19 @@
 module AMS
   module CsvExportExtension
-    CSV_FIELDS = {:GUID=>:id,:Title=>:title,:dates=>:all_dates,:producing_organization=>:producing_organization,
-                  :description=>:description,:level_of_user_access=>:level_of_user_access,:minimally_cataloged=>:minimally_cataloged,
-                  :holding_organization =>:holding_organization_ssim}.freeze
+    CSV_FIELDS = {'asset' =>
+                    { :GUID=>:id,:title=>:title,:dates=>:all_dates,:producing_organization=>:producing_organization, :description=>:description,:level_of_user_access=>:level_of_user_access,:minimally_cataloged=>:minimally_cataloged, :holding_organization =>:holding_organization_ssim}.freeze,
+                  'digital_instantiation' =>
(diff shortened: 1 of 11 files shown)


app/models/ams/csv_export_extension.rb
```diff
@@ -18,16 +24,48 @@ def self.extended(document)
...
+    def export_as_csv(object)
       return CSV.generate do |csv|
-        row = []
-        CSV_FIELDS.each do |csv_field,responder|
+        case object
+        when 'asset'
+          row = []
+          CSV_FIELDS['asset'].each do |csv_field,responder|
+            val = self.send(responder)
+            val = val.join("; ") if val.class == Array
```
afred: maybe better to duck type here with `if val.respond_to? :each`?

app/models/solr_document.rb
```diff
@@ -359,8 +359,12 @@ def playlist_group
...
   end
-  
+
   def media_src(part)
     "/media/#{id}?part=#{part.to_s}"
   end
+
+  def digitized?
+    !sonyci_id.nil? ? sonyci_id.any? : false
+  end
```
afred: this is fine, but you can also do bang bang... i.e. `!!sonyci_id`. This will cast anything to it's boolean equiv. Just gotta be careful with 0, which will cast to true.

app/services/ams/export/documents_to_csv.rb
```diff
@@ -1,14 +1,16 @@
 module AMS
   module Export
     class DocumentsToCsv < ExportService
-      def initialize(solr_documents, format = "csv", filename = nil)
+      def initialize(solr_documents, options={}, format = 'csv', filename = nil)
+        raise "Need to supply an object_type option for CSV export" unless options.has_key?(:object_type)
+        raise "Not a valid object_type for CSV export" unless AMS::CsvExportExtension::CSV_FIELDS.has_key?(options[:object_type])
```
afred: go ahead an `raise ArgumentError` here?

app/models/solr_document.rb
afred: Oh... nevermind.. i guess `sonyci_id` is either nil, an array of stuff, or an empty array?

PR #192, opened 2018-12-21
591 search index design

jasoncorum:
Updates catalog#index view per the ticket. Failing because it needs the solr_wrapper gem update in this PR: #190.

### DIFF
--- a/app/assets/stylesheets/global/search_results.css
+++ b/app/assets/stylesheets/global/search_results.css
@@ -12,3 +12,36 @@ div.facets h3 {
 .search-form .input-group-btn .scope_select {
     border-radius: 0;
 }
+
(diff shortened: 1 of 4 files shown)


afred:
@jasoncorum please review a small change I made in a follow up commit.

jasoncorum:
@afred - thanks for catching that. adding some logic to handle the case where we have an associated collection and the additional col-md-4 is applied.

jasoncorum:
@afred - so i don't think this logic ever gets hit since we're limiting search results to Assets. It's looking for a Collection object for the catalog#index and gets used only when a Collection is in the search results. regardless, i added some logic around our col presentation that matches what gets used for a Collection unless you think it is just better to strip that out entirely?

PR #193, opened 2019-01-08
adding instantiation_admin_data to DigitalInstantiation

adeelahmad:


app/models/digital_instantiation.rb
```diff
@@ -118,6 +119,47 @@ def pbcore_validate_instantiation_xsd
     index.as :stored_searchable, :facetable
   end
 
+  property :instantiation_admin_data_gid, predicate: ::RDF::URI.new('http://pbcore.org#hasInstantiationAdminData'), multiple: false do |index|
+    index.as :symbol
+  end
+
+
+  validates :instantiation_admin_data_gid, presence: true
```
afred: I had some issues with this in `Asset` model where it was validating `admin_data_gid`. Here's the problem...
```
d = DigitalInstantiation.new(location: "foo",
                             digital_format: "bar",
                             media_type: "chu",
                             digital_instantiation_pbcore_xml: File.open('./spec/fixtures/sample_instantiation_valid.xml'))
d.save!
# => ActiveFedora::RecordInvalid: Validation failed: Instantiation admin data gid can't be blank
```

Logically, we're wanting to create an `InstantiationAdminData` record before save, but the validation is happening first. Normally, we would fix this by putting the hook into `before_validation` instead of `before_save`, however, there is a bug in `ActiveFedora` that is causing the `before_validation` hook to not run. There is a fix for it, but it's only in `master`.

For the `Asset` record, I just removed the validation, since the `save_instantiation_admin_data` seems sufficient to guarantee the existence of an associated `InstantiationAdminData` instance.

adeelahmad:
@afred @jasoncorum can either of you guys approve this, i just resolved the conflicts.

PR #194, opened 2019-01-10
More build out of PBCore XML ingest

afred:
**WAIT TO MERGE!** I'm looking into following @adeelahmad's pattern of using existing (default) model actors, rather than creating custom actors... stand by.

* Adds class WGBH::BatchIngest::PBCoreXMLItemIngester to ingest BatchItems
  created by ingest types that read in PBCore XML.
* Changes ingest type to "aapb_pbcore_zipped", with label
  "AAPB PBCore -- Zipped", and uses PBCoreXMLItemIngester.
* Uses Hyrax actor stack when ingesting models. This utilizes a custom actor
  Hyrax::Actors::BatchIngestActor that is swapped in for
  Hyrax::Actors::ModelActor, and works similarly, only instead of calling up
  the defualt Actors for each work type, it calls actors specific for creating
  models from source data submitted via batch ingest.
* Adds actors specific for creating models from data submitted through batch
  ingest, including:
  * Hyrax::Actors::AssetFromPBCoreActor
  * Hyrax::Actors::DigitalInstantiationFromPBCoreActor
* Adds integration spec for aapb_pbcore_zipped ingest type.
* Adds helper methods to make it easy for specs to run an ingest.
* Imports factories for Batch and BatchItem models from hyrax-batch_ingest gem.
* Adds a batch for testing a dozen Asset-only records.
* Adds a helper method zip_to_tmp which will zip a path into a temporary
  location. This is so that we can keep unizpped example batches in the repo,
  but zip them up prior to testing.
* Adds pbcore gem to the bundle.
* Replaces use of PBCore::V2 namespaces from `pb_core` gem with just PBCore
  namespace from `pbcore` gem.

Also,

* Fixes typo in class name from 'Ingestor' to 'Ingester'.
* Fixes typo in dir name from 'btach_ingest' to 'batch_ingest'
* Updates the hyrax-batch_ingest gem in the bundle (along w/ dependencies).
* Adds additional guard clause to avoid errors in AssetActor when dealing with
  typed fields that may not be present.
* Removes validation of Asset#admin_data_gid, this was causing issues due to
  AdminData being created only in a before_save hook (because before_validation
  hook is currently broken in ActiveFedora).
* Fixes bugs in ZippedPBCoreReader and it's spec.
* Fixes bug in before_save hood of Asset model that was not saving AdminData.


afred:
closing in favor of #197

PR #213, opened 2019-02-05
Finalizes player display on asset#show

jasoncorum:
* Stores frame_height and frame_width in DigitalInstantiationActor
* Adds top level empty range for iiif_av
* Only uses Hyrax::IiifAv::ManifestRange if we have media


afred:
@jasoncorum what does this actually change? Can you post a screen shot?

jasoncorum:
The biggest thing display-wise was that there was a 100% width specified in the CSS that we removed and then center the player using an auto margin.

screen shot 2019-02-05 at 10 12 24 am

PR #214, opened 2019-02-05
Changes playback fields on EssenceTrack model

afred:
Replaces playback_frame_per_sec and playback_inch_per_sec with playback_speed
and playback_speed_units.


afred:
### #show view

image

---

### Form view

image

PR #222, opened 2019-02-12
move gems to system

adeelahmad:


afred:
@adeelahmad why are we doing this?

adeelahmad:
@afred since after the upgrade we are no longer using rvm on servers and ruby is installed as system application, now when we use `bundle install --deployment` which installs gems into `./vendor` directory. `Sidekiq` having issues most of the time because of conflicts between gems install into system vs into `./vendor`. This change will make sure, we no longer install gems into `./vendor` but always install gems into system which resolves the conflict.

afred:
@adeelahmad I'm concerned that this will cause issues where `ec2-user` does not have permissions to run gems installed in the system location (owned by root). I've bumped into problems of this nature in the past.

Can you elaborate on which gems are causing the sidekiq problems? I just want t quickly explore a possible alternative than changing permissions on the whole bundle.

PR #227, opened 2019-02-16
646 Maps PBCore ID to Asset ID in XML Ingest

jasoncorum:


app/services/aapb/batch_ingest/pbcore_xml_mapper.rb
```diff
@@ -10,6 +10,8 @@ def initialize(pbcore_xml)
 
       def asset_attributes
         @asset_attributes ||= {}.tap do |attrs|
+          attrs[:id]                          = normalized_id( pbcore.identifiers.select { |id|
+            id.source == "http://americanarchiveinventory.org"}.map(&:value).first)
```
afred: I think we need a conditional around the assignment of `attrs[:id]`, or around the `id.gsub` in `#normalized_id`, or probably both, just to be extra safe. I'd probably lean toward both in this case.

This code needs to be able to handle PBCore XML that does, or does not, have an AAPB ID, and allow ActiveFedora to either use the given ID, or mint a new one.

jasoncorum: 👍

jasoncorum:
One test failed since this branch doesn't have the Faker unique method on the pbcore_title factory that was added in a separate branch prior to the creation of this one. Re-ran the tests.

PR #239, opened 2019-03-01
413 Find Assets from Ingests

jasoncorum:
This adds links to a search page with all Assets associated with an ingest from the batch ingest pages.

### DIFF
--- a/app/presenters/hyrax/asset_presenter.rb
+++ b/app/presenters/hyrax/asset_presenter.rb
@@ -12,8 +12,25 @@ module Hyrax
              :promo_description, :clip_description, :copyright_date,
              :level_of_user_access, :minimally_cataloged, :outside_url, :special_collection, :transcript_status,
              :sonyci_id, :licensing_info, :producing_organization, :series_title, :series_description,
-             :playlist_group, :playlist_order,
+             :playlist_group, :playlist_order, :batch_id,
              to: :solr_document
+
+    def batch
+      raise 'No Batch ID associated with this Asset' unless batch_id.first.present?
+      @batch ||= Hyrax::BatchIngest::Batch.find(batch_id.first)
+    end
+
+    def batch_url
+      @batch_url ||= "/batches/#{batch.id}"
+    end
+
+    def batch_ingest_label
+      @batch_ingest_label ||= Hyrax::BatchIngest.config.ingest_types[batch.ingest_type.to_sym].label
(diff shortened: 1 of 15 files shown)


app/views/hyrax/batch_ingest/batches/index.html.erb
```diff
@@ -0,0 +1,37 @@
...
+          <%= render 'batch', presenter: Hyrax::BatchIngest::BatchPresenter.new(batch) %>
+        <% end %>
+      </tbody>
+    </table>
+    <%= paginate @batches %>
+  </div>
+<% else %>
+  <br />
+  <h4>Batches?!... Batches?!... We don't need no stinking Batches!</h4><br />
+<% end %>
```
afred: @jasoncorum Can we add a comment as to what's being overwritten here? This view comes from `hyrax-batch_ingest` gem, yes?

app/views/hyrax/batch_ingest/batches/show.html.erb
```diff
@@ -0,0 +1,37 @@
...
+        <th scope='col'>Error</th>
+      </tr>
+    </thead>
+    <tbody>
+      <% @presenter.batch_item_presenters.each_with_index do |batch_item_presenter, index| %>
+        <%= render 'batch_item', presenter: batch_item_presenter, index: index %>
+      <% end %>
+    </tbody>
+  </table>
+</div>
```
afred: @jasoncorum same... if this is an overwrite of a view coming from a gem, can we call out specifically what we're overwriting?

jasoncorum:
@afred - i think this covers the items in your review as well as the batch data we display on  asset#show.

app/presenters/hyrax/asset_presenter.rb
```diff
@@ -3,6 +3,8 @@
 module Hyrax
   class AssetPresenter < Hyrax::WorkShowPresenter
 
+    attr_reader :batch, :batch_url, :batch_ingest_label, :batch_ingest_date
```
afred: @jasoncorum these aren't needed if we have explicitly created accessor methods below.

jasoncorum:
Closing this PR without merging. Finally figured out that there was a change in Rails 5 where you now have to opt out of a belongs_to relationship being a required relationship. I have the PBCoreXML working but need to make a few more updates.

PR #244, opened 2019-03-08
Adds links to search results of Assets from Batches

jasoncorum:


app/services/aapb/batch_ingest/csv_item_ingestor.rb
```diff
@@ -26,6 +28,22 @@ def ingest_object_at(node, with_data, with_parent = false)
...
             model_object = node.object_class.constantize.new
+
+            # If ingest is new add batch_id to Asset for tracking
+            if model_object.is_a?(Asset)
+              attributes["hyrax_batch_ingest_batch_id"] = batch_id
+            elsif attributes[:in_works_ids].present?
+              attributes[:in_works_ids].each do |work_id|
+                unless asset = Asset.find(work_id)
+                  raise 'Cannot find Asset with ID: #{work_id}.'
+                end
```
afred: `Asset.find` will raise it's own `ActiveFedora::ObjectNotFoundError` if it can't find the record, so I don't think the custom exception will ever get raised here.

PR #322, opened 2019-06-20
311 batch details say delete

foglabs:


app/views/hyrax/physical_instantiations/_show_actions.html.erb
```diff
@@ -0,0 +1,59 @@
+<%
+# Overwrites Hyrax Gem's version of
+# app/views/hyrax/base/_show_actions.html.erb to explicitly show a
+# download button for Assets that have media available for download.
+
+# Also removes the File Manager link as this feature is not in use
+# in th AMS.
+
+# NOTE: if Hyrax's version of _show_actions.html.erb changes, then this
+# view will have to be updated to reflect those changes if so desired.
```
afred: the comments here don't seem to be accurate. Can we update the comments to reflect what was actually overridden?

PR #328, opened 2019-06-28
325 failure finder

jasoncorum:
@afred - is this all we need at this point for getting the failed batch items? i feel like this might not be exactly what you were looking for.


jasoncorum:
Hold for now. Adding in "expunged" status.

jasoncorum:
If this goes green it should be ready with the 'expunged' status added. i had to where our Gemfile was looking for the hyrax-batch_ingest gem to get the latest there.

jasoncorum:
@afred - Thanks for walking me through stuff on Friday. I think this is what we're looking for now.

PR #331, opened 2019-07-29
Updates to lates hyrax-batch_ingest version

afred:


afred:
# list view
image

# summary view
image

PR #347, opened 2019-09-09
Update README.md

sroosa:
Maybe we should add even more?


sroosa:
Before merging, can we add a brief section about what main software components are required to run AMS (at least in development). Like which versions of Hyrax is works with, etc.

PR #355, opened 2019-09-17
'add view assets link to batch show page'

foglabs:


afred:
@jasoncorum Thanks for the catch. Missed this comment in my inbox, and the PR fell off my radar. But yes, you're right. I failed to add the new file... adding now.

PR #364, opened 2019-09-29
Removes 'pg' gem from bundle

afred:
We needed the 'pg' gem for vaccuuming the PostgreSQL database that Fedora was
using. But we changed Fedora to use MySQL instead, precluding the need for
vacuuming the DB, and the need for the 'pg' gem.

Also,
* Fix indentation in appspec.yml


afred:
Additionally, the 'pg' gem requires some other non-ruby dependencies, which are not currently part of the CloudFormation template used to provision the web instance. I'm trying to get the CloudFormation template working in order to run benchmarks that we want to use for Samvera connect, so this is time sensitive.

afred:
Shamelessly abusing my power to merge this without review. Because I need it.

Issue #375, opened 2019-10-08
No email being sent with exported report or XML


sroosa:
**Describe the bug**
When exporting an Asset Report, a message pops up and says that the report will be emailed to you. After 10 minutes, a report of 30 assets still hasn't been emailed.

**To Reproduce**
1. Go to the batches dashboard
2. Click on View All Asset for Batch 12
3. Click on the Export menu and select the Asset Report option.

**Expected behavior**
A report will be emailed to the user's email account.

**Desktop (please complete the following information):**
Using Chrome on a Mac.

sroosa:
Closed by #380

PR #380, opened 2019-10-16
Adds expected params for csv export & increases the export_to_browser_limit

jasoncorum:


jasoncorum:
Sorry, I forgot to link the ticket on this one. I found the issue while digging in to #375 and if you look at https://github.com/WGBH-MLA/ams/blob/develop/app/services/ams/export/documents_to_csv.rb you'll see that if there are no options passed in then you hit an ArgumentError and this seemed to be the major issue here. Unfortunately this got released without much testing around it.

PR #409, opened 2019-12-11
Corrects solr field name for digital instantiation md5 field

foglabs:


foglabs:
-corrects Solrizer call in `solr_document.rb` for md5 field
-changes to app/views/hyrax/assets/_metadata.html.erb are fixing indentation
-deleted md5 field def on digital instantiation because it was redundant (stored on inst admin data)

PR #421, opened 2020-01-14
'correct search builder for push queries'

foglabs:


app/controllers/pushes_controller.rb
```diff
@@ -52,13 +55,15 @@ def validate_ids
...
         next if index == (requested_ids.count-1)
-        query += %(#{id} OR )
+        query += %(id:#{id} OR )
       end
     end
 
-    query += %(#{requested_ids.last})
-    query_params = {q: query}
-    response, response_documents = search_results(query_params)
+    query += %(id:#{requested_ids.last})
```
afred: Same as above...  I think lines 55-62 can be reduced down to...
```
query += requested_ids.map { |id| "id:#{id}" }.join(" OR ")
```

app/search_builders/ams/push_search_builder.rb
```diff
@@ -4,7 +4,7 @@ class PushSearchBuilder < Hyrax::CatalogSearchBuilder
 
     def leave_fq_alone(solr_parameters)
       solr_parameters[:fq] = [%(has_model_ssim:Asset)]
-      solr_parameters[:q] = %(needs_update:true)
+      # solr_parameters[:q] = %(needs_update:true)
```
afred: remove?

app/services/ams/export/export_service.rb
```diff
@@ -97,7 +97,7 @@ def scp_to_aapb
 
           output << `scp #{aapb_key_path} #{@temp_file_path} ec2-user@#{aapb_host}:/home/ec2-user/ingest_zips/#{@filename}`
           output << `ssh #{aapb_key_path} ec2-user@#{aapb_host} 'unzip -d /home/ec2-user/ingest_zips -o /home/ec2-user/ingest_zips/#{@filename}'`
-          output << `ssh #{aapb_key_path} ec2-user@#{aapb_host} 'cd /var/www/aapb/current && RAILS_ENV=production /home/ec2-user/.gem/ruby/gems/bundler-1.16.5/exe/bundle exec /usr/bin/ruby scripts/download_clean_ingest.rb --files /home/ec2-user/ingest_zips/*.xml'`
+          output << `ssh #{aapb_key_path} ec2-user@#{aapb_host} 'cd /var/www/aapb/current && RAILS_ENV=production /home/ec2-user/.gem/ruby/gems/bundler-1.17.1/exe/bundle exec /usr/bin/ruby scripts/download_clean_ingest.rb --files /home/ec2-user/ingest_zips/*.xml'`
```
afred: You should be able to avoid having to keep updating the path to bundle (which will be required every time bundle increments it's version, even a minor version), by doing the following....
```
# This command does the following...
# 1. SSH's into the AAPB host, using a private SSH key on this host.
# 2. Executes `bash -l` on the AAPB host, which simulate a login so that RVM will work properly, and
#    automatically set the gem path, where the currently installed version of bundler lives.
# 3. Executes the `download_clean_ingest.rb` script in the context of a login bash shell.
output << `ssh #{aapb_key_path} ec2-user@#{aapb_host} 'bash -l -c "cd /var/www/aapb/current && RAILS_ENV=production bundle exec /usr/bin/ruby scripts/download_clean_ingest.rb --files /home/ec2-user/ingest_zips/*.xml"'`
```
This _should_ work as long as the AAPB machine has bash, which it always should. I just tested it from a Rails console on the AMS machine with the following....
```
`ssh -i /home/ec2-user/.ssh/id_rsa ec2-user@americanarchive.org 'bash -l -c "cd /var/www/aapb/current && which bundle"'`
 => "/home/ec2-user/bin/bundle\n" 
```
This shows that by using a `bash -l` then when we remotely `cd` into `/var/www/aapb/current`, it knows about the RVM stuff, enough to select the proper ruby, and gem path (where the installed version of bundler lives).

foglabs: I've had enough different problems with the environment of the ssh session that I'd rather not change this again. I did try the bash -lc route after having problems with the bundle exe not being in my path, but it caused another problem with unzipping

app/controllers/pushes_controller.rb
foglabs: 👍

app/search_builders/ams/push_search_builder.rb
foglabs: 👍

app/controllers/pushes_controller.rb
```diff
@@ -28,13 +28,16 @@ def create
...
       ids.each_with_index do |id, index|
+        # skip the last one
         next if index == (ids.count-1)
-        query += %(#{id} OR )
+        query += %(id: #{id} OR )
       end
     end
-    query += %(#{ids.last})
+
+    query += %(id: #{ids.last})
```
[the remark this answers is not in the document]
foglabs: 👍

PR #478, opened 2020-09-09
Change prod log level from :debug to :warn

afred:
Debug logging is extremely noisy, and I need to track down ingestion errors, so
switching to Warning level for now (shows warnings and errors).


foglabs (approved):
👍

PR #481, opened 2020-09-16
Fixes date formatting for 'Last pushed to AAPB'

### LINKED ISSUE #469: Fix date formatting and/or value for "Last Pushed" on Asset#show page
**Describe the bug**
In some cases, Last Pushed date is either formatted wrongly, has the wrong value, or both.

**To Reproduce**
See record http://ams-edge.wgbh-mla.org/concern/assets/cpb-aacip-db7831c9ad0

The date shown is "01- 1-95 00:00"

**Expected behavior**
Show a readable date. If the data cannot produce a readable date, don't show it only to confuse the user.

**Screenshots**
image

afred:
* Changes #last_pushed and #last_updated methods of AssetPresenter to use
  ApplicationHelper.dispay_date.
* Modifies ApplicationHelper.display_date to:
  * allow input of a different timezone, default to US/Eastern.
  * allow input integer timestamp.
  * logs date formatting errors to Rails log as a WARN before returning nil.
* Change the date formats of #last_pushed and #last_updated to include the
  timezone.

Also,
* refactors specs for AssetPresenter that test for delegated methods; instead of
  calling FactoryBot.create(:asset) 45 times, just leave the Asset out of it and
  use instance_double(SolrDocument) for the one and only dependency into
  AssetPresenter.

Fixes #469.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -20,16 +20,24 @@ module ApplicationHelper
     params
   end
 
-  def display_date(date_time, format: '%Y-%m-%d', from_format: nil)
+  # Module method for displaying dates consistently.
+  # Usage: ApplicationHelper.display_date
+  def self.display_date(date_time, format: '%Y-%m-%d', from_format: nil, time_zone: 'US/Eastern')
     parsed_time = if from_format
-      Date.strptime(date_time, from_format)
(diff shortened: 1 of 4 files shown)


foglabs (approved):
👍

PR #505, opened 2020-11-13
Looking for Sound in Media Player

jasoncorum:
There was a typo introduced to representative_media partial with the media player update. Should check for "Sound" not "Audio" from DigitalInstantiation.


afred (approved):
👍

PR #524, opened 2021-02-26
515 Migration Audit

### LINKED ISSUE #515: AMS1 to AMS2 migration auditing service
## User Story
As an AMS2 website user
I can hit an endpoint on the AMS2 website
Within input parameters:
* `ids` - a list of GUIDs to audit

And get a report of the following information for each GUID:
* Asset exists in AMS2?
  * If exists in AMS2?
  * Number of Digital instantiations in AMS1 vs Number of Digital Instantiations in AMS2
  * For each Digital Instantiation...
    * Number of Essence Tracks in AMS1 vs Number of Essence Tracks in AMS2
  * Number of Physical instantiations in AMS1 vs Number of Digital Instantiations in AMS2

jasoncorum:
Compares the presence of digital instantiations, physical instantiations, and essence tracks present on AMS1 to their corresponding SolDocument on AMS2. Closes #515


app/controllers/audits_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def create
+    ids = params.fetch(:id_field, '').split(/\s+/).reject(&:empty?).uniq
+
+    audit_report = AMS::Migrations::Audit::AuditingService.new(asset_ids: ids, user: current_user).report
+    @matches = audit_report["matches"]
+    @mismatches = audit_report["mismatches"]
+    @errors = audit_report["errors"]
+
+    render 'new'
```
jasoncorum, line 15: redirect_to

app/models/solr_document.rb
```diff
@@ -502,4 +502,28 @@ def md5
...
+    self.member_ids.each do |member_id|
+      all_member_ids << get_members(member_id)
+    end
+    all_member_ids.flatten
+  end
+
+  private
+
+  # Recursively get all members off of members
+  def get_members(id)
```
jasoncorum, line 517: self.get_members

spec/services/ams/migrations/audit/ams1_asset_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+  end
+
+  context 'with a valid AMS 1 ID' do
+
+    before(:each) do
+      allow_any_instance_of(described_class).to receive(:http_response).and_return ams1_response
+    end
+
+    let(:id) { 'cpb-aacip_600-g73707wt6r' }
+    let(:ams1_response) { File.open(File.join(fixture_path, 'batch_ingest', 'sample_pbcore2_xml', 'cpb-aacip_600-g73707wt6r.xml' )).read }
```
jasoncorum, line 50: use pbcore factory to randomize counts

spec/services/ams/migrations/audit/ams2_asset_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'rails_helper'
+require 'pbcore'
+require 'ams/migrations/audit/ams2_asset'
+
+RSpec.describe AMS::Migrations::Audit::AMS2Asset do
+
+  let(:asset) { create(:asset, :with_two_digital_instantiations_and_essence_tracks) }
```
jasoncorum, line 7: let object_count rand[0..10)

app/models/solr_document.rb
```diff
@@ -502,4 +502,28 @@ def md5
...
+    object = SolrDocument.find(id)
+    object.member_ids.each do |member_id|
+      ids << member_id
+      SolrDocument.find(member_id).member_ids.each do |mem_id|
+        get_members(mem_id)
+      end
+    end
+    ids
+  end
+
```
afred, line 528: @jasoncorum on another look, I don't think this recursive method would work as expected. The method returns the `ids` array, but the recursive call on line 523 does not assign the return value to anything. Also, the nested loop means you go from parent-to-grandchild before recursing, but i think we want to recurse at every level.

What about something like this?
```
def self.get_members(id, fam = [])
  fam << id
  id.member_ids.each { |member_id| get_members(member_id, fam) }
  fam
end
```

afred (approved):
👍

PR #533, opened 2021-03-12
Refactors how contributions are added to Asset

afred:
Does not save Asset for every contribution added.
Relies on Hyrax default actor behavior to handle the parent_id, so no need to
explicitly add to ordered members.


afred:
Addition of Bulkrax will obviate this, closing now. Re-open if needed.

Issue #543, opened 2021-04-07
BUG: PBCore XML exports are too slow


afred:
**Describe the bug**
* Exporting a sample of 255 Assets takes several minutes.
* Because 255 is below the browser limit threshold, this means the page sits and spins until the export is complete, which is far too long for users to wait.
* The export size limit for the browser cannot be adjusted per export type, nor do I think we want it to.
* Exports of CSV or PBCore XML shouldn't be that different, and yet the former is literally hundreds, if not thousands, of times faster than the latter.

**To Reproduce**
Do a PBCoreXML export of a set of about 250 records.

**Done when**
* Exports happen in a reasonable time for an end-user to wait in a browser.
* A pbcore xml export of 250 takes less than 10 seconds in a typical AWS stack (requires PR branch deployed to demo stack for test).

PR #548, opened 2021-04-15
Fix slow PBCore XML Export

afred:
* Removes unnecessary Fedora object lookup.
* Adds methods
   * SolrDocument#annotations => returns list of Annotation instances associated with the
     AdminData instance for the record.
   * SolrDocument#admin_data_gid
   * SolrDocument#admin_data => returns AdminData instance
* Replaces module AMS::AllNestedMembersMethods to just AMS::AllMembers.
  * AMS::AllMembers#all_members replaces AMS::AllNestedMembersMethods#all_nested_members.
  * AMS::AllMembers::ClassMethods#get_members (class method mixin) replaces
    instance method mixin AMS::AllNestedMembers#get_members.
* Has instance method `AMS::AllMembers#all_members` call recursive class method, and remove `self` because
  it is confusing when `self` is part of `self#all_members`.
* Use return value to accumulate recursive results SolrDocument.get_members instead of pass-by-ref
  array param.
* Add complimentary SolrDocument#all_members similar to AMS::AllMembers#self, but returns SolrDocument
  instances instead of ActiveFedora::Base instances.
* Have #all_members return objects instead of just IDs to avoid requiring additional lookups when used.
  NOTE: we have to fetch the objects anyway to get the child IDs, so might as well return them and then
  map to IDs if needed.
* Adds filtering parameters to SolrDocument#members for returning/excluding member records
* Replaces SolrDocument#find_child(klass) with SolrDocument#members(only: klass).
  for specific model types.
* Changes SolrDocument#members to not run n queries; 1 query for all members.
* NOTE: AMS::Solr::CreateMembersMethods#find_child was doing this already, but now we
  are memoizing the result, and have a quick return if no member_ids.


foglabs (approved):

afred:
Closes #543

PR #554, opened 2021-04-23
Fixes deployment info page

afred:
The problem was that the script trying to fetch deployment info was using AWS credentials
for the user who has permissions on S3, but not CodeDeploy.

This uses a different set of credentials which it expects to be correctly set in the
environment. If the expected env vars are empty, the error message will say say.


afred:
I tested this on the demo server, since it only runs after deployment on the deployed host.

http://ec2-3-85-189-114.compute-1.amazonaws.com/deployment.html

image

foglabs (approved):

PR #555, opened 2021-04-30
Adds PBCore XML endpoint

afred:
When on an Asset page, add .xml to the GUID and get the PBCore XML back.

Also,
* Removes redundant inclusion of Hydra::Controller::ControllerBehavior in CatalogController


jasoncorum (approved):
👍

PR #559, opened 2021-05-04
Completes Validate and Clean for PBCore

jasoncorum:
This PR includes a custom validation class for a PBCore Description Document (from the WGBH pbcore gem) that checks for certain types, values, and elements that (while they may be technically valid in PBCore) are not exactly what we want for OUR PBCore. The PR also includes a Cleaner that checks the PBCore for similar characteristics and edits the PBCore Description Document.


afred (approved):
👍

PR #568, opened 2021-05-10
Fixes SolrDocumet#members

afred:
The 'rows' parameter was defaulting to 10, which was limiting the number of children
returned, which was causing instantiations to not be returned for PBCore XML exports
and pushes to AAPB. This fixes that, by setting the 'param' to arbitrarily high
number. Also, sets rows param for #member_of query.


jasoncorum (approved):
💥

PR #581, opened 2021-05-18
Do not add empty Contributions

### LINKED ISSUE #579: BUG: Do not include empty Contribution columns on CSV ingest
**Describe the bug**
* Found an example of a record with a bunch of empty Contributions associated with it.
* Tracked it back to Batch 574, BatchItem 206386
* The empty Contribution records come from empty contribution data: each having a single Contributor that is just an empty string, and that's all.
* This makes it's way all the way to AAPB

**To Reproduce**
Have yet to reproduce locally, but my guess is that if you do a CSV ingest where one row has a bunch of empty Contribution columns, then this same problem will result.

**Expected behavior**
Empty Contributor columns on a CSV ingest do not result in empty Contribution records in AMS.

afred:
* Fixes CsvReader to filter out empty values when mapping to BatchItem#source_data JSON.
* Adds a CSV factory following this example: https://id-in-it.com/factorybot-for-csv-file-fixtures
  * CsvBuilder is a thin wrapper around CSV ruby lib that simplifies the interface and adds a
    #save method that FactoryBot needs for `create`

Fixes #579.

### DIFF
--- a/app/services/aapb/batch_ingest/csv_reader.rb
+++ b/app/services/aapb/batch_ingest/csv_reader.rb
@@ -25,7 +25,7 @@ module AAPB
       end
 
       def validate_options
-        @options_structure = AAPB::BatchIngest::CSVConfigParser.validate_config(@options)
+        @options_structure = AAPB::BatchIngest::CSVConfigParser.validate_config(options)
       end
 
       def validate_csv_header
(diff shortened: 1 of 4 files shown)


jasoncorum (approved):
💥

PR #587, opened 2021-06-04
Adds pry to all environments

### LINKED ISSUE #586: BUG: Failing ingets with "Only referenceable nodes may be the value of reference properties"
**Describe the bug**
See batch: http://ams2.wgbh-mla.org/batches/952?locale=en (And other batches trying to ingest `Peabody_stragglersmigration.zip`)

**To Reproduce**
I created a script to run ingests from Rails console on Demo, and have tried running an ingest using a pared down version of `Peabody_stragglersmigration.zip` (having only 1 file), and was able to reproduce.

**Expected behavior**
Either the thing succeeds at ingest, or there is a better error about why it went wrong, and what to do about it.

**Background:**
Old notes 11/19/2020 in Slack (#cool-room) suggest this has something to do with `Ldp::Gone`, or some similar violation of a uniqueness constraint in the RDB.

**Done when**
1. Identify the cause
2. file new ticket based on the cause

afred:
The error "Only referenceable nodes may be the value of reference properties" is a somewhat cryptic way to say that you're trying to save a new object with an ID that has a tombstone associated with it (i.e. it existed before in the repository, but has since been deleted, but not fully eradicated).

Eradicating the tombstone for these IDs allows us to move forward with the ingest. So that's good, but once again, I found myself really wanting a pry session on our demo machine, so I'm just adding it to the production environment here, for that reason. Debugging in prod env is not ideal, but sometimes you gotta do it to flush out issues specific to the prod environment, and I've had to do that several times now.

closes #586

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -42,10 +42,6 @@ end
 group :development, :test do
   # Use sqlite3 as the database for Active Record
   gem 'sqlite3', '1.3.13'
-  # Call 'byebug' anywhere in the code to stop execution and get a debugger console
-  gem 'byebug'
-  gem 'pry-byebug', platforms: [:mri, :mingw, :x64_mingw]
-  gem 'pry-rails'
   gem 'capybara-screenshot'
   gem 'rspec', "~> 3.7"
   gem 'rspec-rails', "~> 3.7"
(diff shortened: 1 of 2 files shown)


jasoncorum (approved):
👍

PR #590, opened 2021-06-08
Removes IIIF Player Methods 

jasoncorum:
Some records are failing to load because of an error when the asset_presenter attempts to calculate duration from an EssenceTrack. The only reason the seconds were needed were for the IIIF Player that we recently replaced with VideoJS. This PR removes the duration method as well as other methods and checks that were needed to satisfy the requirements of the IIIF Player.


foglabs (approved):
👍

PR #593, opened 2021-07-09
2265 Proxy Time Start Annotation

jasoncorum:
This adds "Proxy Time Start" as an Annotation type and adds a few missing tests from the Asset model.


afred (approved):
👍

afred (approved):
👍

PR #599, opened 2021-07-23
Allows updating of an Assets organization annotation

### LINKED ISSUE #598: Allow adding of organization annotations to existing Assets via batch
### Because
@mivillesvik has the following use case:
```
As an archivist
I can do an ingest of type `AAPB CSV Asset Attribute Update Ingester`
with CSV headers: `Asset, Asset.id, Asset.organization`
and it will save an organization annotation to the Asset.
``` 

### Done when
above user story is satisfied

afred:
Closes #598.

### DIFF
--- a/config/batch_ingest.yml
+++ b/config/batch_ingest.yml
@@ -83,6 +83,7 @@ ingest_types:
           - licensing_info
           - playlist_group
           - playlist_order
+          - organization
   aapb_csv_reader_4:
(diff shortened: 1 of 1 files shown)


afred:
stand by for screenshot of passing local specs.

afred:
image
in lieu of Travis build ☝️

jasoncorum (approved):
👍

PR #604, opened 2021-08-13
Get Sony Ci ID from GUID via UI

### LINKED ISSUE #585: Add capability to get Filename from Sony Ci ID
### Because
We need to be able to map from GUID to Sony Ci ID

This will be used in API endpoint specified in #584, and can be POSTed to via webhook from Sony Ci after files are uploaded.

### Done when
1. We can get Asset data from Sony Ci (including the filename) from a Sony Ci ID, using the Sony Ci API

afred:
* Updates sony_ci_api gem to use sony_ci_api_rewrite in github, branch v0.1.
	* Had to create a branch that was compatible with Faraday v0.17.x, b/c Hyrax dependencies
	  would not allow Faraday v1.3.
* Customizes Sony Ci ID input to include filename fetched from Sony Ci.
* Customizes Sony Ci ID input partial to include button that will populate Sony Ci ID
  entries from Sony Ci, where the filename matches a translated form of the Asset ID.
* Replaces old instances of SonyCiBasic with SonyCiApi::Client.
* Adds a SonyCi::APIController to wrap very specific requests to Sony Ci API using the
  new gem. Endpoints include:
  * /sony_ci/api/find_media&query={QUERY}&fields={FIELDS}
  	This will search the API for records matching {QUERY} and return JSON objects having properties
	specified by the comma-separated field list {FIELDS}
  * /sony_ci/api/get_filename&sony_ci_id={SONY_CI_ID}
    This will find the Sony Ci record and return a JSON object containing fields 'sony_ci_id' and 'name'.

Closes #585.


afred:
Here's the screenshot of the passing specs from my local machine until we get CI up again.
image

PR #614, opened 2021-09-01
Refactors error handling in MediaController#show

### LINKED ISSUE #609: BUG: MediaController#download_url does not fail gracefully
**Describe the bug**
If the response from Sony Ci happens to be nil for whatever reason, then you'll get Rails's default 500 HTML NoMethoError, which absolutely destroys the JS that is trying to consume it.

**To Reproduce**
1. On Demo, configure Sony Ci to point to the Sandbox workspace (i.e. the one with barely any media in it). This ensures that the existing records on Demo will not be able to find their Sony Ci media.
2. Try to access a #show or #edit interface for an Asset, where the Sony Ci ID of the assets is _not_ in the Sandbox workspace on Sony Ci.
3. Look in the JS Console and the Network tab (on Google Chrome Inspector), and you can see the 500 error, and the HTML response.

**Expected behavior**
If a `SonyCiApi::Cilent` returns an HTTP error status, then MediaController#download_url should return the same status, along with an empty string.

**Actual Behavior**
The `MediaController#download_url` is assuming that the response returns a hash, and when it doesn't, it's causing a NoMethodError, which always results in a 500 http response from the controller action... so the real error is not getting properly handled, and thus falsely masked.

**Screenshots**
Here's the errors in the Console.
image

afred:
Some error cases were not being covered:
* when SonyCiApi gem returns a SonyCiApi error.
* when the SolrDocument is missing the sonyci_id_ssim field.
* when the sonyci_id_ssim field is there, but empty.

Some of these errors were resulting in ugly 500 errors with HTML rendering. But this controller
is specific to redirect to a Sony Ci URL where the media is, and return the byte stream of the
media, so HTML reponses are out.

Here we always return the redirect on success, and a head response (no body) with the appropriate
HTTP status on failure.

closes #609.

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -1,24 +1,38 @@
 require 'sony_ci_api'
 
 class MediaController < ApplicationController
+
+  rescue_from StandardError, with: :error_response_default
+  rescue_from Blacklight::Exceptions::RecordNotFound, with: :error_response_404
+  rescue_from SonyCiApi::HttpError, with: :error_response_from_sony_ci
+
   def show
-    if solr_document
-      if can? :show, solr_document
-        redirect_to download_url
-      else
-        head :forbidden
-      end
-    else
-      head :not_found
-    end
+    head(:forbidden) and return unless can? :show, solr_document
+    head(:not_found) and return unless download_url
+    redirect_to download_url and return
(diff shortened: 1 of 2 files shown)


jasoncorum (approved):
👍

PR #615, opened 2021-09-02
Fix 'Find media in Sony Ci...' button

### LINKED ISSUE #608: BUG: "Find Sony Ci records..." button does not work on first page load.
**To Reproduce**
1. Go to the Edit interface for an Asset
2. Click on the button for "Find all Sony Ci files matching [GUID]"

**Expected behavior**
1. The text below the button says "searching" briefly
2. The text blow is replaced with "[N] records found"
3. The Sony CI IDs are copied to the input fields below when N > 0.

**Actual Behavior**
It refreshes the page. The expected behavior described above then works after the first page refresh.

**Background**
I think this might be a turbolinks issue. The JS events are attached to the DOM elements in a `$(document).on 'turbolinks:load'` event handler, but before it runs, the href for the link is `#`. Clicking on the button then, for some reason, refreshes the page.

If you go to the Edit Asset page, and _do not_ click the "Find all Sony Ci Files matching.... " button, but simply refresh the page, then the button will work as expected. So actually clicking the button has nothing to do with it.

Also, I don't think I encountered this in my dev environment, so it may be a prod env issue as well.

afred:
This button (as a <a> tag currently) is rendered on the Asset Edit interface in the Admin Data
section of the form, and finds Sony Ci media whose filename matches an Asset's GUID. But it
wasn't working on initial page. The behavior was being added in the turbolinks:load event,
which should work (we're doing it elsewhere), but on this page it simply would not. So I
changed it to use the jQuery(document).ready() method, which typically does NOT work because
of turbolinks, but in this case, it works better. But still not sure why.

Closes #608.

### DIFF
--- a/app/assets/javascripts/sony_ci/find_media.coffee
+++ b/app/assets/javascripts/sony_ci/find_media.coffee
@@ -85,9 +85,12 @@ class FindSonyCiMediaBehavior
     $(@sonyCiIdInputSelector).change @fetchFilenameHandler
 
 # When the page loads...
-$(document).on 'turbolinks:load', ->
+# NOTE: could not get $(document).on('turbolinks:load') to work on initial page
+# load; reverting to $(document).ready, which seems to work more consistently.
+$(document).ready ->
   # This regex matches the 3rd URL segment which should be the GUID.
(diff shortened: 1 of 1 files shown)


jasoncorum (approved):
👍

PR #618, opened 2021-09-09
Fixes Cartrivision + 28mm

### LINKED ISSUE #519: Physical Instantiation Cartrivision misspelling
## User Story
In the Physical Instantiation "Formtas" drop-down menu, Cartrivision is spelled "Cartivision". According to the internet it looks like the format name wants that r in there! (I attached screenshots below)

## Done Looks Like
"Cartivision" would be spelled "Cartrivision" in the physical instantiation format drop-down menu and would map correctly upon ingest with physical instantiation fields naming "Cartrivision" as the format.

jasoncorum:
Fixes misspelling of "Cartrivision" and the term for "28mm film."

Closes #519

### DIFF
--- a/config/authorities/format.yml
+++ b/config/authorities/format.yml
@@ -138,7 +138,7 @@ terms:
   - id: 22mm film
     term: 22mm film
   - id: 28mm film
-  - term: 28mm film
+    term: 28mm film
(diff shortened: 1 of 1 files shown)


afred (approved):
👍

PR #621, opened 2021-09-10
Re-indexes Asset after webhook from Sony Ci

### LINKED ISSUE #620: BUG: sony_ci/webhooks/save_sony_ci_id is not re-indexing the Solr doc
### User story
As a Sony Ci media manager
When the Sony Ci webhook exist and correctly configured to POST to `sony_ci/webhooks/save_sony_ci_id` on event `AssetProcessingFinished`
And after I upload a media file with name `{guid}.mp4`
And after that media file is finished processing
The new Sony Ci ID is saved to the Asset record in Fedora
And the Sony Ci ID is re-indexed to the Asset's solr document.
In order to link an AMS Asset with a Sony Ci media file (also referred to as an Asset).

### Actual behavior
The Sony Ci ID _is_ saved to the Asset, but it is _not_ saved to the Solr document.

### Background
The code is only calling `AdminData#save`, but not calling `Asset#save`, which would re-index the Asset's solr document with the new Sony Ci ID field.

### Done when
user story is true

afred:
Re-indexes Asset after webhook from Sony Ci

The Asset was not being reindexed, and so it was saving on the Asset's admin data, but not the Solr doc.

This explicitly calls Asset#save! to re-index, but Asset#save! also persists in Fedora, which is slow, so if
there's a faster way to just save AdminData and re-index without Fedora, we should do that

Closes #620.

### DIFF
--- a/app/controllers/sony_ci/webhooks_controller.rb
+++ b/app/controllers/sony_ci/webhooks_controller.rb
@@ -16,14 +16,23 @@ module SonyCi
     end
 
     def save_sony_ci_id
-      asset_admin_data_from_sony_ci_filename.update!( sonyci_id: [ sony_ci_id ] )
-      render status: 200, json: { message: "success" }
+      asset.admin_data.update!( sonyci_id: [ sony_ci_id ] )
+      # Re-save the Asset to re-index it.
+      # TODO: Is there a faster way to save the Sony Ci ID to the AdminData and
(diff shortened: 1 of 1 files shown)


afred:
image

afred:
also... 29 minutes!!! 😬

jasoncorum (approved):
⭐

PR #656, opened 2022-04-16
queue based indexing during import, Add rbspy to docker image

orangewolf:


package.json
```diff
@@ -2,10 +2,11 @@
   "name": "AMS",
   "private": true,
   "engines": {
-    "node": "12.22.10"
```
orangewolf, line 5: @alishaevn is this going to cause problems?

config/application.rb
```diff
@@ -40,5 +40,17 @@ class Application < Rails::Application
 
     # Don't generate system test files.
     config.generators.system_tests = nil
+
+    config.to_prepare do
```
orangewolf, line 44: this makes autoloading _decorator.rb files automatic in both dev and prod mode.

app/jobs/bulkrax/child_relationships_job_decorator.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
```
orangewolf, line 1: @alishaevn is this supposed to be in here?

afred (approved):
👍

package.json
alishaevn: I don't believe so. we changed it because it was causing the github actions in this repo to fail.

talked about it [here](https://assaydepot.slack.com/archives/C030UPFFP2A/p1646684584156969?thread_ts=1646684579.748519&cid=C030UPFFP2A).

app/jobs/bulkrax/child_relationships_job_decorator.rb
alishaevn: I probably copied it over from another decorator so it may be able to be removed. but it hasn't caused me any issues that I can recall at the moment.

alishaevn:
closing in favor of https://github.com/WGBH-MLA/ams/pull/658

PR #664, opened 2022-06-30
'explicitly enforce dash guid format for thumbnail files from AAPB'

foglabs:


mrharpo (approved):
Would love to put a period on this, but I ... :disguised_face: must :dash:!

PR #670, opened 2022-07-15
Use setup-node action for installing node

afred:
Replaces a shell command for installing node in order to get the proper version installed correctly.


mrharpo (approved):
:+1: Simple and sweet! :robot:

PR #672, opened 2022-07-15
Allows batch update of plain title and description

### LINKED ISSUE #667: Plain Title/ Plain Description not allowed to batch-update in AMS2
**Describe the bug**
When trying to run a batch-update in AMS2 for plain title (Asset.title) and plain description (Asset.description) the AMS2 just spits it out and gives a header error. These headers work upon ingest, but only upon update do these fail.

**To Reproduce**
1. make a spreadsheet with GUIDs already in AMS2
2. add columns for Asset.title and/or Asset.description
3. attempt to ingest as a batch

**Expected behavior**
Ideally, the AMS2 would allow these headers to be updated!

**Screenshots**

afred:
Previously you could only update specific title and description types, e.g. program_title,
series_description. This adds the non-typed (plain) Assett.title and Asset.description as allowed
CSV headers in a batch update.

Closes #667.

### DIFF
--- a/config/batch_ingest.yml
+++ b/config/batch_ingest.yml
@@ -59,12 +59,14 @@ ingest_types:
           - eidr_id
           - topics
           - subject
+          - title
           - program_title
           - episode_title
           - segment_title
           - raw_footage_title
           - promo_title
           - clip_title
+          - description
           - program_description
           - episode_description
           - segment_description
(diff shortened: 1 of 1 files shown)


mrharpo (approved):
:+1: Looks good!

ekemeyer:
Tested - there's still some wonkiness, maybe because of how titles are stored in Fedora?
For example, if you use asset update ingester, it performs an add-on to the existing title - it does not replace.

PR #676, opened 2022-08-30
Merge to ams

orangewolf:


.env
```diff
@@ -0,0 +1,39 @@
+APP_NAME=gbh-hyrax
+CHROME_HOSTNAME=chrome
+DB_ADAPTER=mysql2
+FCREPO_URL=http://fcrepo:8080/rest
+MYSQL_DATABASE=gbh
+MYSQL_HOST=db
+MYSQL_PASSWORD=DatabaseFTW
+MYSQL_ROOT_PASSWORD=DatabaseFTW
```
afred, line 8: ☝️ do these passwords ever work?

.env
```diff
@@ -0,0 +1,39 @@
...
+MYSQL_USER=root
+REDIS_HOST=redis
+REDIS_SERVER=redis
+SETTINGS__BULKRAX__ENABLED=true
+SIDEKIQ_CONCURRENCY=1
+SOLR_ADMIN_PASSWORD=SolrRocks
```
afred, line 14: keep password?

.github/workflows/deploy-staging.yaml
```diff
@@ -0,0 +1,38 @@
+name: "Deploy Staging"
```
afred, line 1: We'll only have `ams-production` and `ams-demo` namespaces in kubernetes, so idk if we need a workflow for a "staging" deployment.

.sops.yaml
```diff
@@ -0,0 +1,3 @@
+---
+creation_rules:
+  - pgp: "B6125B16B0DD59F34D6975FBF885927FDA9C48E2"
```
afred, line 3: is this a secret?

Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+
+USER app
+
+RUN mkdir -p /app/fits && \
+    cd /app/fits && \
+    wget https://github.com/harvard-lts/fits/releases/download/1.5.0/fits-1.5.0.zip -O fits.zip && \
+    unzip fits.zip && \
+    rm fits.zip && \
+    chmod a+x /app/fits/fits.sh
+ENV PATH="${PATH}:/app/fits"
```
afred, line 40: Do we still want fits?

.gitignore
```diff
@@ -47,3 +48,10 @@ demo*
 
 # This file is generated on deployment
 public/deployment.html
+
+# Ignore vendor files
+/vendor/*
+.dory.yml
+solr_db_initialized
+docker-compose.override.yml
+*~undo-tree~
```
afred, line 59: i'm not sure what each of these are for:
```
.dory.yml
solr_db_initialized
docker-compose.override.yml
*~undo-tree~
```

.github/workflows/deploy-production.yaml
```diff
@@ -0,0 +1,37 @@
...
+jobs:
+  deployment:
+    runs-on: 'ubuntu-latest'
+    container: dtzar/helm-kubectl:3.9.4
+    env:
+      DEPLOY_IMAGE: ghcr.io/${{ github.repository }}
+      WORKER_IMAGE: ghcr.io/${{ github.repository }}/worker
+      HELM_EXPERIMENTAL_OCI: 1
+      HELM_RELEASE_NAME: gbh-production
+      KUBE_NAMESPACE: gbh-production
```
afred: This probably needs to change to `ams-production`, yes?

.github/workflows/deploy-production.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+jobs:
+  deployment:
+    runs-on: 'ubuntu-latest'
+    container: dtzar/helm-kubectl:3.9.4
+    env:
+      DEPLOY_IMAGE: ghcr.io/${{ github.repository }}
+      WORKER_IMAGE: ghcr.io/${{ github.repository }}/worker
+      HELM_EXPERIMENTAL_OCI: 1
+      HELM_RELEASE_NAME: gbh-production
```
afred: Is `HELM_RELEASE_NAME` supposed to match the kube namespace? If so, it probably needs to be `ams-production`

.github/workflows/deploy-production.yaml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Set env
+        run: >-
+          echo "TAG=${GITHUB_SHA::8}" >> $GITHUB_ENV;
+          echo "DEPLOY_TAG=${GITHUB_SHA::8}" >> $GITHUB_ENV;
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Do deploy
+        run: >-
+          echo $KUBECONFIG_FILE | base64 -d > $KUBECONFIG;
+          DOLLAR=$ envsubst < ops/production-deploy.tmpl.yaml > ops/production-deploy.yaml;
```
afred, line 35: I only see `ops/staging-deploy.tmpl.yaml`. Won't we need `ops/demo-deploy.tmpl.yaml` and `ops/production-deploy.tmpl.yaml`.

.github/workflows/deploy-production.yaml
```diff
@@ -0,0 +1,37 @@
...
+          cat ops/production-deploy.yaml;
+          ./bin/helm_deploy gbh-production gbh-production
```
afred: should this match the kube namespace or the helm release name?

.github/workflows/deploy-demo.yaml
```diff
@@ -0,0 +1,37 @@
...
+          echo "TAG=${GITHUB_SHA::8}" >> $GITHUB_ENV;
+          echo "DEPLOY_TAG=${GITHUB_SHA::8}" >> $GITHUB_ENV;
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Do deploy
+        run: >-
+          echo $KUBECONFIG_FILE | base64 -d > $KUBECONFIG;
+          DOLLAR=$ envsubst < ops/demo-deploy.tmpl.yaml > ops/demo-deploy.yaml;
+          cat ops/demo-deploy.yaml;
+          ./bin/helm_deploy gbh-demo gbh-demo
```
afred: see comments for deploy-production.yaml

Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+
+COPY --chown=1001:101 $APP_PATH/Gemfile* /app/samvera/hyrax-webapp/
+RUN bundle install --jobs "$(nproc)"
+
+COPY --chown=1001:101 $APP_PATH /app/samvera/hyrax-webapp
+
+ARG SETTINGS__BULKRAX__ENABLED="false"
+RUN sh -l -c " \
+  yarn install && \
+  RAILS_ENV=production SECRET_KEY_BASE=fake-key-for-asset-building-only DB_ADAPTER=nulldb bundle exec rake assets:precompile"
```
afred, line 50: are these values used for building the production image? If so, can we parameterize them? If not, where do the production values come from?

README.md
```diff
@@ -1,14 +1,122 @@
...
+
+# Docker development setup
+
+We recommend committing .env to your repo with good defaults. .env.development, .env.production etc can be used for local overrides and should not be in the repo. See [Handling Secrets with SOPS](#handling-secrets-with-sops) for how to manage secrets.
+
+1) Install Docker.app
+
+2) Install stack car
+    ``` bash
+    gem install stack_car
```
afred, line 19: need an explainer for stack_car

README.md
```diff
@@ -1,14 +1,122 @@
...
+    ```
+
+3) Sign in with dory
+    ``` bash
+    dory up
```
afred, line 24: how is dory being used and how it's being configured?

README.md
```diff
@@ -1,14 +1,122 @@
...
+In terminal:
+```bash
+bin/encrypt-secrets
+```
+
+This will find and output an encrypted version of secret files with an `.enc` extension.
+
+Release and Deployment are handled by the gitlab ci by default. See ops/deploy-app to deploy from locally, but note all Rancher install pull the currently tagged registry image
+
+## Staging Deploys: N8 Architecture
```
afred, line 84: replace this section with WGBH deployments?

app/actors/hyrax/actors/asset_actor.rb
```diff
@@ -13,6 +13,8 @@ def create(env)
         add_description_types(env)
         add_date_types(env)
 
+        # queue indexing if we are importing
+        env.curation_concern.reindex_extent = "queue#{env.importing.id}" if env.importing
```
afred, line 17: i need a refresher on what setting the `reindex_extent` actually does.

app/actors/hyrax/actors/environment.rb
```diff
@@ -0,0 +1,25 @@
...
+  module Actors
+    class Environment
+      # @param [ActiveFedora::Base] curation_concern work to operate on
+      # @param [Ability] current_ability the authorizations of the acting user
+      # @param [ActionController::Parameters] attributes user provided form attributes
+      def initialize(curation_concern, current_ability, attributes, importing = false)
+        @curation_concern = curation_concern
+        @current_ability = current_ability
+        @attributes = attributes.to_h.with_indifferent_access
+        @importing = importing
```
afred, line 14: Is the importing flag set from inside Bulkrax then?

app/controllers/hyrax/batch_ingest/batches_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+  def create
+    # OVERRIDE HYRAX-Batch_Ingest revision: dc9d38039728eab581ab7b1cb55cf9ff33984b13
+    redirect_to '/importers/new'
+  end
+
+  def index
+    # OVERRIDE HYRAX-Batch_Ingest revision: dc9d38039728eab581ab7b1cb55cf9ff33984b13
+    redirect_to '/importers'
+  end
+end
```
afred, line 35: Could this be wrapped in the conditional for turning Bulkrax on/off?

app/controllers/hyrax/my_controller.rb
```diff
@@ -19,6 +19,7 @@ def self.configure_facets
         config.add_facet_field IndexesWorkflow.suppressed_field, helper_method: :suppressed_to_status
         config.add_facet_field solr_name("resource_type", :facetable), limit: 5
         config.add_facet_field solr_name("hyrax_batch_ingest_batch_id", :stored_searchable)
+        config.add_facet_field solr_name("bulkrax_importer_id", :stored_searchable)
```
afred, line 22: is `:stored_searchable` a `_tesim` field? Not sure how much it matters.

app/factories/bulkrax/object_factory.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  # Regardless of what the Parser gives us, these are the properties we are prepared to accept.
+  def permitted_attributes
+    klass.properties.keys.map(&:to_sym) + %i[id edit_users edit_groups read_groups visibility work_members_attributes admin_set_id member_of_collections_attributes pbcore_xml skip_file_upload_validation bulkrax_importer_id]
+  end
+
+  def environment(attrs)
+    Hyrax::Actors::Environment.new(object, Ability.new(@user), attrs, @importer)
+  end
+end
```
afred, line 27: Is this an override of something in Bulkrax, and if so, what is being overridden?

app/indexers/asset_indexer.rb
```diff
@@ -17,6 +17,7 @@ def generate_solr_document
       solr_doc['broadcast_date_drsim'] = object.broadcast_date if object.broadcast_date
       solr_doc['created_date_drsim'] = object.created_date if object.created_date
       solr_doc['copyright_date_drsim'] = object.copyright_date if object.copyright_date
+      solr_doc[Solrizer.solr_name('bulkrax_identifier', :facetable)] = object.bulkrax_identifier
```
afred, line 20: Into what is `bulkrax_identifier` a foreign key?

config/initializers/redis_config.rb
```diff
@@ -1,3 +1,9 @@
...
+sentinels = config[:sentinel] && config[:sentinel][:host].present? ? { sentinels: [config[:sentinel]] } : {}
+redis_config = config.except(:sentinel).merge(thread_safe: true).merge(sentinels)
 require 'redis'
-config = YAML.safe_load(ERB.new(IO.read(Rails.root.join('config', 'redis.yml'))).result)[Rails.env].with_indifferent_access
-Redis.current = Redis.new(config.merge(thread_safe: true))
+Redis.current = begin
+                  Redis.new(redis_config)
+                rescue
+                  nil
+                end
```
afred: when would we want to use sentinel?

config/initializers/hyrax.rb
```diff
@@ -231,10 +246,20 @@
...
 Date::DATE_FORMATS[:standard] = "%m/%d/%Y"
 
 Qa::Authorities::Local.register_subauthority('subjects', 'Qa::Authorities::Local::TableBasedAuthority')
 Qa::Authorities::Local.register_subauthority('languages', 'Qa::Authorities::Local::TableBasedAuthority')
 Qa::Authorities::Local.register_subauthority('genres', 'Qa::Authorities::Local::TableBasedAuthority')
+
+# set bulkrax default work type to first curation_concern if it isn't already set
+if ENV['SETTINGS__BULKRAX__ENABLED'] == 'true'
+  if Bulkrax.default_work_type.blank?
+    Bulkrax.default_work_type = Hyrax.config.curation_concerns.first.to_s
```
afred: should this always be `Asset`?

app/indexers/queued_nesting_indexer.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class QueuedNestingIndexer
+  extend Samvera::NestingIndexer
+  def self.reindex_relationships(id:, maximum_nesting_depth: nil, extent:)
+    if extent.match("queue")
+      Rails.logger.info("nested indexing queued")
+      Redis.current.zadd("nested:index:#{extent.delete("queue")}", 0, id.to_s)
```
afred, line 8: Here it looks like we're identifying whether the extent is of the form "queueN", and if so, then adding the `id` to a sorted list in Redis at key "nested:index:N", with a score of 0.

And from code above, it looks like all our model's are having their `reindex_extent` set to "queueN" where "N" is the `env.importing_id`.

Is `env.importing_id` then the `bulkrax_identifier`?

How does this end up looking in Redis, and how does that end up informing the nested indexing?

config/initializers/hyrax.rb
```diff
@@ -191,7 +191,22 @@
...
+        old_url = "#{ActiveFedora.fedora.host}#{ActiveFedora.fedora.base_path}/#{::Noid::Rails.treeify(id, false)}"
+        ActiveFedora::Fedora.instance.connection.head(old_url)
+      rescue Ldp::NotFound
+        # Do nothing
+      else
+        url = old_url
+      end
+    end
+    url
+  end
```
afred, line 209: Need to make sure that this isn't conflicting with logic from `IdentifierService`. I don't think it is, but want verification.

app/jobs/bulkrax/child_relationships_job_decorator.rb
```diff
@@ -0,0 +1,66 @@
...
+        work_membership
+      end
+      # Not all of the Works/Collections exist yet; reschedule
+    rescue Bulkrax::ChildWorksError
+      # OVERRIDE bulkrax v.1.0.0
+      # In case the work hasn't been created, don't endlessly reschedule the job
+      attempts = (args[3] || 0) + 1
+      child_ids = @missing_entry_ids.presence || args[1]
+
+      reschedule(args[0], child_ids, args[2], attempts) unless attempts > 5
```
afred, line 23: can we think of any cases where we wouldn't want to reschedule in the event of `ChildWorksError`? I.e. in cases where it's not waiting for records to get saved?

And if the max attempts is exceeded, does it mean the records are saved, but the child relationships are never created? If so, what is the remedy?

app/jobs/bulkrax/index_after_job.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    def perform(importer)
+      # check if importer is done, otherwise reschedule
+      pending_num = importer.entries.left_outer_joins(:latest_status)
+        .where('bulkrax_statuses.status_message IS NULL ').count
+      return reschedule(importer.id) unless pending_num.zero?
+
+      # read queue and index objects
+      set = Redis.current.zpopmax("nested:index:#{importer.id}", 100)
+      logger.error(set.to_s)
```
afred: `logger.info` here instead?

app/jobs/bulkrax/index_after_job.rb
```diff
@@ -0,0 +1,30 @@
...
+      return if set.blank?
+      loop do
+        set.each do |key, score|
+          Hyrax.config.nested_relationship_reindexer.call(id: key, extent: 'full')
+        end
+        set = Redis.current.zpopmax("nested:index:#{importer.id}", 100)
+        logger.error(set.to_s)
```
afred: `logger.info` here instead?

app/jobs/bulkrax/importer_job.rb
```diff
@@ -0,0 +1,21 @@
...
+  Bulkrax::ImporterJob.class_eval do
+    def perform(importer_id, only_updates_since_last_import = false)
+      importer = Bulkrax::Importer.find(importer_id)
+
+      importer.current_run
+      unzip_imported_file(importer.parser)
+      import(importer, only_updates_since_last_import)
+      update_current_run_counters(importer)
+      schedule(importer) if importer.schedulable?
+      Bulkrax::IndexAfterJob.set(wait: 1.minute).perform_later(importer)
```
afred, line 16: add comment above this line indicating it's part of the override.

Also, could this be moved to an `after_perform` callback? I notice that in the `IndexAfterJob#perform` there is logic to check the importer to see if it's finished, and if not, it reschedules itself. Wondering if that rescheduling could go away if the `IndexAfterJob` is enqueued in `after_perform` of the `ImporterJob`, that way we know it's done.

app/models/bulkrax/csv_entry.rb
```diff
@@ -0,0 +1,162 @@
...
+      if value.is_a? Hash
+        flatten_hash(value, hash, index)
+      else
+        index.present? ? hash["#{key}_#{index}"] = value : hash[key] = value
+      end
+
+      hash
+    end
+  end
+end
```
afred, line 162: This seems to map a CSV row to all the various models, so I'm interested in a bit of a walkthrough on the logic and where it fits into the Bulkrax import process.

app/models/bulkrax/pbcore_xml_entry.rb
```diff
@@ -0,0 +1,105 @@
...
+          source: annotation['source'],
+          value: annotation['value'],
+          annotation: annotation['annotation'],
+          version: annotation['version'],
+          admin_data_id: admin_data.id
+        )
+      end
+    end
+  end
+end
```
afred, line 105: This seems to map a PBCore document to all the various models, so I'm interested in a bit of a walkthrough on the logic and where it fits into the Bulkrax import process.

app/models/concerns/bulkrax/has_matchers.rb
```diff
@@ -0,0 +1,58 @@
...
+      end
+    when Hash
+      Array.wrap(content)
+    when String
+      Array.wrap(content.strip)
+    else
+      Array.wrap(content)
+    end
+  end
+end
```
afred, line 58: Could use a bit of clarification on this. It's matching Bulkrax import column labels to object fields, or something like that?

spec/services/ams/media_download/media_download_service_spec.rb
```diff
@@ -41,7 +41,7 @@
           result = service.process
           expect(result).to be_a(AMS::MediaDownload::MediaDownloadService::Success)
           expect(result[:filename]).to match(/(export-)\d{2}_\d{2}_\d{4}_\d{2}:\d{2}.zip/)
-          expect(result[:file_path].path).to match(/(export-)\d{2}_\d{2}_\d{4}_\d{2}:\d{2}.zip/)
+          expect(result[:file_path].path).to match(/(export-)\d{2}_\d{2}_\d{4}_\d{4}.zip[\d-]+/)
```
afred, line 44: Was this failing? It looks like the `@filename` for `MediaDownloadService` has not changed.

spec/rails_helper.rb
```diff
@@ -39,6 +39,46 @@
...
+  Capybara.register_driver :chrome do |app|
+    Capybara::Selenium::Driver.new(
+      app,
+      browser: :chrome,
+    )
+  end
+end
+
+Capybara.javascript_driver = :chrome
+
```
afred, line 81: Was the remote driver necessary for some circumstances? When and why should we use it?

spec/models/bulkrax/csv_entry_spec.rb
```diff
@@ -0,0 +1,202 @@
...
+        end
+      end
+
+      describe 'writes the csv file' do
+        before do
+          exporter.write
+        end
+
+        xit 'with the asset and its children in a single row' do
+          # TODO(alishaevn): write this spec
```
afred, line 196: TODO?

afred: Also, is this where we would put in a round-tripping test for:
1. export as CSV
2. edit
3. re-import

app/jobs/bulkrax/delete_work_job.rb
```diff
@@ -0,0 +1,17 @@
...
+    work = entry.factory.find
+    if work.is_a? Asset
+      asset_destroyer = AMS::AssetDestroyer.new
+      asset_destroyer.destroy([work.id])
+    end
+    importer_run.increment!(:deleted_records)
+    importer_run.decrement!(:enqueued_records)
+  end
+  # rubocop:enable Rails/SkipsModelValidations
+end
```
afred, line 17: I'm a little fuzzy on this. Is this called for non-Assets? Looks like it will increment/decrement whether or not anything was actually deleted, which does not seem right.

afred (commented):
NOTE: This is a partial review.

.env
orangewolf: this password is used for local dev env if doing docker based set up. they are never meant to be secure

.env
orangewolf: this password is used for local dev env if doing docker based set up. they are never meant to be secure

.github/workflows/deploy-staging.yaml
orangewolf: this is the N8 staging. we can delete this workflow if we stop using that environment.

.sops.yaml
orangewolf: its a gpg id - the secrete is the private part of the gpg key, which you either have already or will get so you can decrypt some of the provisioning stuff that _is_ secret.

.gitignore
orangewolf: - .dory.yml and solr_db_initialized are not relevent. I'll pull them out. 
- docker-compose.override.yml is a common pattern in docker-compose. it allows local dev copies to customize their setup w/o effecting the basic dev set up.
- *~undo-tree~ is an artifact of Spacemacs, which is the text editor I use. basically they are undo files that let me do complex undo and to keep undo on a file between editing sessions.

.github/workflows/deploy-production.yaml
orangewolf: yep, I'll fix it

.github/workflows/deploy-production.yaml
orangewolf: they don't _have_ to but we are only deploying this one app to the namespace, so they may as well.

.github/workflows/deploy-production.yaml
orangewolf: yep. fixing

.env
afred: 👍

.github/workflows/deploy-demo.yaml
orangewolf: 👍

.env
afred: 👍

.github/workflows/deploy-staging.yaml
afred: 👍

.sops.yaml
afred: 👍

.gitignore
afred: 👍

Dockerfile
orangewolf: these values are ONLY used during the asset build. It doesn't need the secret key base or the database to make the assets, but Rails sprockets implementation boots the app in order to do the asset building (one of the many reasons it takes soooo long). So fake values for the db and secret key are fine here. The assets are still "good" when the image is deployed.  The SECRET_KEY_BASE and the various database config are injected as ENV vars in to the Kubernetes pod (aka the running container).  Those values come from the ops/deploy template and get "real" values from the secrets stored in the CI workflow.

Dockerfile
orangewolf: as long as no one every uploads any files, then no. if we have it in the image and someone does upload an image at least it will work. I left it mostly becauseI thought it was installed on your cloud-formation generated machines.

README.md
orangewolf: stack car is a wrapper around docker-compose that we use to provide some opinionated docker-compose defaults. `rm -rf tmp/pids/* && docker-compose up web` becomes `sc up`. We can easily update the readme to not include it if you'd prefer.

README.md
orangewolf: dory is a proxy for docker containers running locally. Basically it allows you to connect to running containers on your machine w/o binding all of the ports. it also makes doing ssl simple in dev mode by handling the self signed cert. https://github.com/FreedomBen/dory

README.md
orangewolf: yep, though, I think that's the other PR

app/actors/hyrax/actors/environment.rb
orangewolf: yep, this is how it survives the actor stack.

app/actors/hyrax/actors/asset_actor.rb
orangewolf: reindex_extent is usually "limited" or "full". It is used by the nesting indexer to handle whether children (and descendants) should be reindexed every time this object is reindexed or not. We use it here to define a 3rd option which is to queue the child reindexing.  This queueing allows us to solve this problem:

add item A to the collection
- item A is indexed
- collection is indexed
add item B to item A
- item B is indexed
- item A is indexed
- collection is indexed
add item C to the collection
- item C is indexed
- collection is indexed
- item A is indexed
- item B is indexed
add item D to the collection
- item D is indexed
- collection is indexed
- item A is indexed
- item B is indexed
- item C is indexed

etc. 

We instead queue the collection and parent indexing and do it one time at the end of the import run.

app/controllers/hyrax/batch_ingest/batches_controller.rb
orangewolf: easy enough

app/controllers/hyrax/my_controller.rb
orangewolf: yes, it does make it a tesim.

app/factories/bulkrax/object_factory.rb
orangewolf: this is allowing for the passing of the importer "flag" to the actor stack.  The need for the actor stack is for the queued indexing extent implementation. That's not in Bulkrax or Hyrax yet, because we architected it here in GBH to solve the indexing speed issue.

app/indexers/asset_indexer.rb
orangewolf: the bulkrax_identifier ties the Bulkrax::Entry object to the curation concern (aka ActiveFedora::Base descendent) record it creates.

config/initializers/redis_config.rb
orangewolf: no, I do not think it is needed. It was part of the Hyrax chart at one point, but isn't any more. We do still need the rescue because the app won't load w/o a redis connection. we want to be able to load the app w/o services present to create assets (see Dockerfile comments above).

config/initializers/hyrax.rb
orangewolf: yes.

app/indexers/queued_nesting_indexer.rb
orangewolf: importing_id is the id of the Importer, not of the individual Import Entry.  So for "Test Import 18", the id might be 40. We queue all the records from that import in to "nested:index:40" and then pull them during IndexAfterJob in sets of 100 for batch conversion to Solr. By indexing 100 records at a time in Solr instead of 1 record at a time, we cut the number of http requests for Solr and make indexing more efficient ON TOP of the fact that we solved the N+1 indexing issue by queuing them in the first place. This implementation was informed by Searchkick, and Ruby gem for Elastic Search which has a similar queued indexing method.

config/initializers/hyrax.rb
orangewolf: absolutely feel free to do your own independent verification, but my understanding is that `IdentifierService` mints new UUIDs using secure random, and then checks them in Solr. It only handle the id part and not the full url. This code only handles the fedora document url part, leaving the id part at the end intact in all cases. Or at least that was the architecture goal.

app/jobs/bulkrax/child_relationships_job_decorator.rb
orangewolf: These jobs are queued at the end of the line so to speak. So the only time child jobs can start to run is if all the create jobs are run, errored or are in the process of running. If they are run, we are golden. If they are errored, this job times out, but the import entry is marked as a failure and this job would get automatically rerun when an import of that work is successful. if the job is still running - that's when this retry logic comes in to play. This basically says if you started creating a child _and_ you've been working on it for more than 15 minutes, we think its actually errored and not really running.

app/jobs/bulkrax/index_after_job.rb
orangewolf: probably logger.debug - I'll update it

app/jobs/bulkrax/importer_job.rb
orangewolf: I agree about the comment. I'll add it.

we'd still need the retry logic because its not just after this job, but after all the import jobs for a given import set.

app/models/bulkrax/csv_entry.rb
orangewolf: yep, we can do that. This is the code @ShanaLMoore wrote to take a single row with multiple objects in it and map that to a Bulkrax::Entry for each object.

app/models/bulkrax/pbcore_xml_entry.rb
orangewolf: yep, we can do that. This is the code @ShanaLMoore wrote to take a single row with multiple objects in it and map that to a Bulkrax::Entry for each object.

spec/services/ams/media_download/media_download_service_spec.rb
orangewolf: yes, it was failing, it was the only failing spec when we set up our CI process.

spec/rails_helper.rb
orangewolf: remote driver is needed if you want to do dockerized spec runs and not have to install chrome (and all it many dependencies) in the the docker image. Since we use docker for our dev and the CI uses it for running the specs, having chrome remote means we don't have to ship chrome in the production image, but we can still use it for testing. It drops roughly 1GB off the size of the image, so thats worth it.

Dockerfile
afred: 👍

Dockerfile
afred: 👍

app/jobs/bulkrax/delete_work_job.rb
orangewolf: what the counter should count is kinda tricky here. The CSV should only have assets in it. The requirement was to specifically skip any non-assets marked with delete. BUT a row that has a delete but is not an assets is still successfully processed so it theoretically should count toward the total processed row amount.

.env
orangewolf: lol - just found a pending review in github where I said "this password is used for local dev env if doing docker based set up. they are never meant to be secure" - guess that would have been helpful if I'd hit the right button on it

app/actors/hyrax/actors/asset_actor.rb
afred: 👍 mainly wondering what Hyrax expected values are.

app/controllers/hyrax/batch_ingest/batches_controller.rb
afred: 👍

spec/models/bulkrax/csv_entry_spec.rb
orangewolf: @alishaevn can you shed some light on this one?

app/controllers/hyrax/my_controller.rb
afred: I seem to recall having had problems when storing IDs in `_tesim` fields because Solr would tokenize. But I don't recall the specifics, i.e. if it was due to the ID format or what. In any event, I set a flag in my brain to try to store IDs as symbols (e.g. ssim) rather than text.

Like I said, not sure if it matters here. If not, then 👍

app/factories/bulkrax/object_factory.rb
afred: 👍

config/initializers/redis_config.rb
afred: 👍

spec/models/bulkrax/csv_entry_spec.rb
alishaevn: I believe I initially left this as a placeholder to show that a parent asset and it's child works would all be exported on a single row. however, we're not actually creating a csv in this spec and I believe the describe block above for "creates the correct headers" is verification enough.

so this can be removed. suggestion below.

spec/models/bulkrax/csv_entry_spec.rb
```diff
@@ -0,0 +1,202 @@
...
+        end
```
alishaevn, line 197:
```suggestion
```

alishaevn: ref: https://github.com/WGBH-MLA/ams/pull/676/files#r964130070

app/indexers/queued_nesting_indexer.rb
afred: 👍

.github/workflows/deploy-production.yaml
afred: 👍

.github/workflows/deploy-production.yaml
afred: 👍

.github/workflows/deploy-production.yaml
afred: 👍

README.md
afred: 👍 that's fine, just wondering where it might show up.

README.md
afred: 👍

app/indexers/asset_indexer.rb
afred: 👍

config/initializers/hyrax.rb
afred: 👍

app/jobs/bulkrax/child_relationships_job_decorator.rb
afred: 👍

app/jobs/bulkrax/importer_job.rb
afred: 👍

app/models/bulkrax/csv_entry.rb
afred: 👍

config/initializers/hyrax.rb
afred: I think I'd prefer if it was just explicit then, i.e. `Bulkrax.default_work_type = "Asset"`, because I can't tell whether `Bulkrax.default_work_type` has already been set, and if not, then I'm not sure how we are guaranteed that `Hyrax.config.curation_concerns.first` is `Asset`.

spec/rails_helper.rb
afred: 👍

app/jobs/bulkrax/delete_work_job.rb
afred: 👍

config/database.yml
```diff
@@ -1,28 +1,25 @@
...
-  # The time that MySQL2 lib will try to get a db connection.
-  # Default is 5.
-  # checkout_timeout: 10
-  pool: 12
+  <<: *login
+
+test:
+  <<: *login
+  database: <%= ENV['TEST_DB'] %>
+
```
afred, line 19: 👇 remove staging config?

docker-compose.yml
```diff
@@ -0,0 +1,233 @@
...
+        condition: service_started
+      fcrepo:
+        condition: service_started
+      redis:
+        condition: service_started
+      zoo:
+        condition: service_started
+      initialize_app:
+        condition: service_completed_successfully
+
```
afred, line 219: 👇 is this `chrome` service name what must match `CHROME_HOSTNAME`?

solr/security.json
```diff
@@ -0,0 +1,15 @@
...
+      "realm":"My Solr users",
+      "forwardCredentials": false
+  },
+  "authorization":{
+      "class":"solr.RuleBasedAuthorizationPlugin",
+      "permissions":[{"name":"security-edit",
+                      "role":"admin"}],
+      "user-role":{"solr":"admin"}
+  }
+}
```
afred, line 15: this is completely new to me, so I could use a little primer on why it's now needed, where these values are used, and whether they need to match up with any other config values elsewhere.

spec/features/search_by_dates_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'rails_helper'
 
-RSpec.describe 'Search by dates', reset_data: false, js: true do
+RSpec.describe 'Search by dates', reset_data: false, js: true, skip: true do
```
afred, line 3: I assume we're skipping this because it started failing. Do we know why? Date range search is a feature we want to make sure is still working.

spec/fixtures/bulkrax/xml/deleted.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<pbcoreDescriptionDocument xmlns="http://www.pbcore.org/PBCore/PBCoreNamespace.html" xmlns:xsi="xsi" xsi:schemaLocation="http://www.pbcore.org/PBCore/PBCoreNamespace.html http://www.pbcore.org/xsd/pbcore-2.0.xsd" xsi:xmlns="http://www.w3.org/2001/XMLSchema-instance">
+  <pbcoreAssetDate dateType="created">1996-05-11</pbcoreAssetDate>
+  <pbcoreIdentifier source="http://americanarchiveinventory.org">cpb-aacip/20-000000hr</pbcoreIdentifier>
+  <delete>true</delete>
```
afred, line 5: I don't recall, is this a way to tell Bulkrax to delete the record if it's found?

solr/config/schema.xml
```diff
@@ -308,7 +308,7 @@
     <dynamicField name="*_fsi" type="float" stored="true" indexed="true" multiValued="false"/>
     <dynamicField name="*_fsim" type="float" stored="true" indexed="true" multiValued="true"/>
 
-    <!-- trie float (_ft...) (for faster range queries) -->
+    <!-- FloatPointField (_ft...) (for faster range queries) -->
     <dynamicField name="*_fti" type="tfloat" stored="false" indexed="true" multiValued="false"/>
     <dynamicField name="*_ftim" type="tfloat" stored="false" indexed="true" multiValued="true"/>
     <dynamicField name="*_fts" type="tfloat" stored="true" indexed="false" multiValued="false"/>
```
afred, line 314: I assume these changes are due to the Solr upgrade. Without digging into the specifics of each field type, is there any change in Solr's behavior that we should be aware of?

lib/ams.rb
```diff
@@ -64,7 +64,7 @@ def run_migrations!
       end
 
       def flush_redis_cache!
-        Redis.new(host: ENV.fetch('REDIS_SERVER', 'localhost'), port: '6379').flushall
+        Redis.new(host: ENV.fetch('REDIS_HOST', 'localhost'), port: '6379').flushall
       end
```
afred, line 68: I'm still seeing references to `REDIS_SERVER` in:
* `.env`
* `config/cable.yml`
* `ops/staging-deploy.tmpl.yaml`
Can we get rid of references to `REDIS_SERVER` altogether?

ops/staging-deploy.tmpl.yaml
```diff
@@ -0,0 +1,178 @@
...
+solr:
+  enabled: false
+
+externalFcrepoHost: fcrepo.staging-fcrepo.svc.cluster.local
+
+solrExistingSecret: solr-admin-password
+externalSolrHost: solr.staging-solr.svc.cluster.local
+externalSolrUser: admin
+externalSolrCollection: "gbh-hyrax-staging"
+externalSolrPassword: $SOLR_PASSWORD
```
afred, line 178: See other comment, but i'm not sure what the desired end-result is here. Seems like we should have `demo-deploy.tmpl.yaml` and `production-deploy.tmpl.yaml` and remove `staging-deploy.tmpl.yaml`

hyrax/templates/NOTES.txt
```diff
@@ -0,0 +1,21 @@
...
+{{- else if contains "LoadBalancer" .Values.service.type }}
+     NOTE: It may take a few minutes for the LoadBalancer IP to be available.
+           You can watch the status of by running 'kubectl get --namespace {{ .Release.Namespace }} svc -w {{ include "hyrax.fullname" . }}'
+  export SERVICE_IP=$(kubectl get svc --namespace {{ .Release.Namespace }} {{ include "hyrax.fullname" . }} --template "{{"{{ range (index .status.loadBalancer.ingress 0) }}{{.}}{{ end }}"}}")
+  echo http://$SERVICE_IP:{{ .Values.service.port }}
+{{- else if contains "ClusterIP" .Values.service.type }}
+  export POD_NAME=$(kubectl get pods --namespace {{ .Release.Namespace }} -l "app.kubernetes.io/name={{ include "hyrax.name" . }},app.kubernetes.io/instance={{ .Release.Name }}" -o jsonpath="{.items[0].metadata.name}")
+  echo "Visit http://127.0.0.1:8080 to use your application"
+  kubectl --namespace {{ .Release.Namespace }} port-forward $POD_NAME 8080:3000
+{{- end }}
```
afred, line 21: I'd like to do a high-level overview of what all the files under hyrax/templates and hyrax/charts are responsible for and how they are summoned to work.

hyrax/.helmignore
```diff
@@ -0,0 +1,23 @@
...
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
```
afred, line 23: does this work along with `.dockerignore`?

config/bulkrax_api.yml
```diff
@@ -0,0 +1,84 @@
...
+        - "Create"
+        - "Create and Import"
+        - "Update Importer"
+        - "Update and Re-Import (update metadata and replace files)"
+        - "Update and Harvest Updated Items', 'Update and Re-Harvest All Items"
+        - "Update and Re-Import (update metadata only)"
+        - "Update and Import (importer has not yet been run)"
+    auth_token:
+      definition: 'Authentication token. Required for JSON requests only.'
+      required: true
```
afred, line 84: where do these values end up?

config/application.rb
```diff
@@ -37,5 +40,17 @@ class Application < Rails::Application
...
+      # Allows us to use decorator files, which change methods or behavior on upstream classes
+      # with minimal overrides or fuss. Pattern adapted from Spree and Refinery projects
+      Dir.glob(File.join(File.dirname(__FILE__), "../app/**/*_decorator*.rb")).sort.each do |c|
+        Rails.configuration.cache_classes ? require(c) : load(c)
+      end
+
+      Dir.glob(File.join(File.dirname(__FILE__), "../lib/**/*_decorator*.rb")).sort.each do |c|
+        Rails.configuration.cache_classes ? require(c) : load(c)
+      end
+    end
```
afred, line 54: Is this a result of running a Hyrax or Bulkrax generator? I don't see any decorators in `app/decorators/` or `lib/decorators/` that AMS is using.

bin/solrcloud-upload-configset.sh
```diff
@@ -0,0 +1,42 @@
...
+      echo "--   It's possible SolrCloud is bootstrapping its configuration, so this process will retry."
+      echo "--   see: https://solr.apache.org/guide/8_6/configsets-api.html#configsets-upload"
+    fi
+  fi
+  COUNTER=$(( COUNTER+1 ));
+  sleep 5s
+done
+
+echo "--- ERROR: failed to create Solr ConfigSet after 5 minutes";
+exit 1
```
afred, line 42: I assume this is part of the Solr upgrade. Is there any GBH/AMS specific configuration here I we should know about?

app/views/hyrax/dashboard/sidebar/_repository_content.html.erb
```diff
@@ -12,8 +12,20 @@
...
-    <%= menu.nav_link(hyrax_batch_ingest.batches_path) do %>
-      <span class="fa fa-table" aria-hidden="true"></span> <span class="sidebar-action-text"><%= 'Batches' %></span>
+  <% if ENV['SETTINGS__BULKRAX__ENABLED'] == 'false' %>
+    <% if can? :index, Hyrax::BatchIngest::Batch %>
+      <%= menu.nav_link(hyrax_batch_ingest.batches_path) do %>
+        <span class="fa fa-table" aria-hidden="true"></span> <span class="sidebar-action-text"><%= 'Batches' %></span>
+      <% end %>
     <% end %>
   <% end %>
+
```
afred: 👇 Does this need to be inside the conditional for `ENV['SETTINGS__BULKRAX__ENABLED']`?

app/services/aapb/batch_ingest/pbcore_xml_mapper.rb
```diff
@@ -44,8 +44,16 @@ def prepare_annotations(annotations)
...
+
+        if ENV['SETTINGS__BULKRAX__ENABLED'] == 'true'
+          type_id
+        else
+          if type_id.present?
+            type_id
+          else
+            raise "annotation_type not registered with the AnnotationTypesService: #{type}."
+          end
+        end
```
afred, line 56: This logic indicates that if we're using Bulkrax, but the annotation type is not found then `type_id` will be `nil` rather than raising an error.

Do we know what happens when `type_id` is `nil`? Does it raise an error elsewhere?

.github/workflows/deploy-production.yaml
orangewolf: yes, i was leaving creating these as an exercise to do together since we'll need info from the fcrepo / solr deploy to fill them out

config/initializers/hyrax.rb
orangewolf: 👍 already changed it

.env
orangewolf: @ShanaLMoore can you provide some clarification on this?

.github/workflows/deploy-staging.yaml
orangewolf: yes

.sops.yaml
orangewolf: in our standardized set up, we secure solr with a simple user/password. The big reason for this the collection and config apis are not available unless you have password enabled. You'll see we use solr cloud here, mostly because we need it for many of our projects. This file configures solr with a simple solr:SolrRocks user/password pair for development.

Dockerfile
orangewolf: this one failed when we started. Once we get merged it is probably a good time to pull the skip off and see what's going on.

.gitignore
orangewolf: yes

solr/config/schema.xml
orangewolf: that's correct. it seems to act the same as far as I can tell from Ruby land. It is more about how things are handled inside solr.

lib/ams.rb
orangewolf: REDIS_SERVER is the "whole" thing `redis://localhost:6379/1`where REDIS_HOST is just the domain part (`localhost` or `redis`). We could rewrite things to not need REDIS_SERVER, but I think you end up adding several variables to accomplish that, so its probably not worth while.  Also the hyrax helm chart uses REDIS_SERVER so you get stuck with it there not matter what.

hyrax/templates/NOTES.txt
orangewolf: this directory is the helm chart from Hyrax, modified to use mysql instead of postgres. Helm chart overviews can found here: https://helm.sh, https://www.freecodecamp.org/news/what-is-a-helm-chart-tutorial-for-kubernetes-beginners/ 


Big broad strokes is that the template files get combined with the passend in values from ENV.tmpl.yaml (in ops) to produce the yaml files that are sent to the K8S cluster and declare state. This is similar to how a Rails view is a template that has variables filled in to create an HTML file. 

So to deploy the worker you need to create a deployment-worker.yaml file that has the all values for that environment filled in to it (to know where solr and redis are for example, and to have the db password). Helm does the templeting and creates versions of the rolled out yaml files which allows for roll backs.

hyrax/.helmignore
orangewolf: it prevents files from ending up in the helm packages sent to the cluster.

.github/workflows/deploy-production.yaml
orangewolf: this file, currently in its boilerplate form, defines how the bulkrax CLI is used.

.github/workflows/deploy-production.yaml
orangewolf: interpret_visibility_actor and child_relationships_job. We use this decorator pattern whenever we override a file from a gem dependency. This makes it easier to target the exact change we want instead of copying the whole file over.

.github/workflows/deploy-production.yaml
orangewolf: it copies the config from `solr/config` in to the dockerized dev environment.

.github/workflows/deploy-production.yaml
orangewolf: yes, seems like there should be an else clause

.github/workflows/deploy-demo.yaml
orangewolf: this was one of those validation errors that Tim keeps talking about. A lot of the example data sets would fail on this error, but the records seem to be otherwise valid. I thought that the goal was to let the ams1 stuff in close to as is at this point. This may need requirement discussion?

config/bulkrax_api.yml
orangewolf: this file, currently in its boilerplate form, defines how the bulkrax CLI is used.

config/application.rb
orangewolf: interpret_visibility_actor and child_relationships_job. We use this decorator pattern whenever we override a file from a gem dependency. This makes it easier to target the exact change we want instead of copying the whole file over.

app/services/aapb/batch_ingest/pbcore_xml_mapper.rb
orangewolf: this was one of those validation errors that Tim keeps talking about. A lot of the example data sets would fail on this error, but the records seem to be otherwise valid. I thought that the goal was to let the ams1 stuff in close to as is at this point. This may need requirement discussion?

app/views/hyrax/dashboard/sidebar/_repository_content.html.erb
orangewolf: yes, looks like there should be an else clause

bin/solrcloud-upload-configset.sh
orangewolf: it copies the config from solr/config in to the dockerized dev environment.

spec/fixtures/bulkrax/xml/deleted.xml
orangewolf: yes

app/models/concerns/bulkrax/has_matchers.rb
orangewolf: @ShanaLMoore can you provide some clarification on what this code change is doing?

docker-compose.yml
orangewolf: yes, thats how it gets found inside docker

spec/features/search_by_dates_spec.rb
orangewolf: It was failing when we started. I recommend we get things merged and then pull the skip out and see if it is still flapping or not.

solr/security.json
orangewolf: basically solr requires a username/pass to access the config and collection APIs. we use these to set up solr in a consistent way in docker across hyrax/hyku projects. When you docker-compose down it clears everything and then docker-compose up it loads the config from solr/config and makes a new collection using that config.

orangewolf: because it is just development env, we set it to solr:SolrRocks here.

afred (approved):
👍

PR #680, opened 2022-10-22
move to gbh libs aws account, deploy for alpha-delta servers, k8s provisioning

orangewolf:
This PR is in a few parts:

1) deploy script has been consolidated to take an argument
`MYSQL_PASSWORD=VALUE ./bin/deploy.sh ENV SHA` for example `MYSQL_PASSWORD=VALUE ./bin/deploy.sh prod 4068fd0e`

2) ops templates for alpha/bravo/charlie/delta

3) modified prod and demo templates because new cluster uses efs-sc instead of efs for EFS PVCs and uses the internal ips of the fcrepo/solr server instead of the public ips

4) provisioning for stuff in the k8s cluster. Specifically cert-manager and the efs storage class.


ops/alpha-deploy.tmpl.yaml
```diff
@@ -0,0 +1,137 @@
...
+    mountPath: /app/samvera/hyrax-webapp/public/uploads
+    subPath: public-uploads
+  - name: uploads
+    mountPath: /app/samvera/hyrax-webapp/tmp/network_files
+    subPath: network-files
+
+ingress:
+  enabled: true
+  hosts:
+    - host: ams2-alpha.wgbh-mla.org
```
afred, line 36: I'm not sure this has been applied. The ingress named `ams-alpha-hyrax` seems to have host `ams2.wgbh-mla.org`, which is the production host

ops/alpha-deploy.tmpl.yaml
```diff
@@ -0,0 +1,137 @@
...
+  enabled: false
+
+brandingVolume:
+  storageClass: efs-sc
+derivativesVolume:
+  storageClass: efs-sc
+uploadsVolume:
+  storageClass: efs-sc
+
+extraVolumeMounts: &volMounts
```
afred, line 16: 👇 why are all of these named the same thing, and why does it look like the `subPath` and `mountPath` are redundant?

ops/prod-deploy.tmpl.yaml
```diff
@@ -51,11 +51,11 @@ extraEnvVars: &envVars
   - name: FCREPO_BASE_PATH
     value: /prod
   - name: FCREPO_HOST
-    value: 52.90.209.247:8080
+    value: 10.0.4.93:8080
```
afred, line 54: ☝️ why are we switching from public to private here?

ops/provision/modules/ec2/main.tf
```diff
@@ -27,15 +27,19 @@ resource "aws_instance" "fcrepo" {
...
     prod = "prod"
     demo = "demo"
+    alpha = "alpha"
+    bravo = "bravo"
+    charlie = "charlie"
+    delta = "delta"
   }
 
-  ami                         = data.aws_ami.amazon-linux-2.id
+  ami                         = "ami-02538f8925e3aa27a" # data.aws_ami.amazon-linux-2.id
```
afred, line 36: ☝️ why this change? Do we need to pin to this AMI?

ops/provision/modules/k8s/files/prod_issuer.yaml
```diff
@@ -0,0 +1,19 @@
...
+kind: ClusterIssuer
+metadata:
+  name: letsencrypt-prod
+  namespace: cert-manager
+spec:
+  acme:
+    # The ACME server URL
+    server: https://acme-v02.api.letsencrypt.org/directory
+    # Email address used for ACME registration
+    email: support@notch8.com
```
afred, line 11: Not sure what this process looks like, but did we want to switch this from a `notch8.com` email?

ops/provision/modules/k8s/main.tf
```diff
@@ -0,0 +1,90 @@
...
+}
+
+provider "kubectl" {
+  config_path = var.kubeconfig
+}
+
+provider "kubernetes" {
+  config_path = var.kubeconfig
+}
+
```
afred, line 28: 👇 I don't see this resource, nor a `ingress-nginx` namespace. Is this yet to be applied, or no longer used, or am I misunderstanding something?

ops/provision/variables.tf
```diff
@@ -46,3 +46,13 @@ variable "site24x7_key" {
...
+
+variable "kubeconfig" {
+  type = string
+  default = "./kube_config.yaml"
+}
+
+variable "efs_name" {
+  type = string
+  default = "fs-0dd9f8ff037001c5d"
+}
```
afred, line 58: ☝️ are these redundant with variables defined in ops/provision/modules/k8s/variables.tf?

ops/provision/modules/k8s/main.tf
```diff
@@ -0,0 +1,90 @@
...
+  namespace = "ingress-nginx"
+  create_namespace = true
+  version = "4.3.0"
+  repository = "https://kubernetes.github.io/ingress-nginx"
+  chart = "ingress-nginx"
+  values = [
+    file("modules/k8s/files/ingress-nginx-values.yaml")
+  ]
+}
+
```
afred, line 40: 👇 i'm unsure what role this resource plays. Also not seeing a namespace for "storage",  so maybe this hasn't been applied, or maybe i'm missing it?

ops/provision/modules/networking/security-groups.tf
```diff
@@ -12,6 +12,35 @@ resource "aws_security_group" "access" {
...
+    description      = "AMS2 on digital-eks-dev cluster"
+    from_port        = 8983
+    ipv6_cidr_blocks = []
+    prefix_list_ids  = []
+    protocol         = "tcp"
+    security_groups  = []
+    self             = false
+    to_port          = 8983
+  }
+
```
afred, line 43: ☝️ these ingresses aren't needed if demo and prod are within `r2-gbh-ams` cluster, yeah?

ops/alpha-deploy.tmpl.yaml
orangewolf: I fixed this during code review, so yes, it needs to be pushed up.

ops/alpha-deploy.tmpl.yaml
orangewolf: the name is which mount volume from above is used (uploads for most of these). then subPath is what subdirectory on the mounted volume to use and mountPath is where in the container to mount it. For example
```
  - name: uploads
     mountPath: /app/samvera/hyrax-webapp/tmp/network_files
     subPath: network-files
```

Says "on the uploads volume make a subdirectory called network-files. then mount that subdirectory to  /app/samvera/hyrax-webapp/tmp/network_files in the running container"

ops/prod-deploy.tmpl.yaml
orangewolf: they are in the same aws availability zone / account now so internal traffic is a) free and b) faster c) not requiring security group configuration

ops/provision/modules/ec2/main.tf
orangewolf: the AMI got a new release and I didn't want to blow away and rebuild demo/prod while I was hustling to get this work done. once we've got the rush over, rebuild is fairly simple and relatively safe.

ops/provision/modules/k8s/files/prod_issuer.yaml
orangewolf: This email is only used if the renew fails. it should probably be y'all. Updating it is just changing it here and then deploying.

ops/provision/modules/k8s/main.tf
orangewolf: this is the actual "hook up efs to the eks cluster" code. I see it in the cluster... not sure why you do not.

ops/provision/modules/k8s/main.tf
orangewolf: its deployed. I see it 
Screen Shot 2022-10-24 at 14 22 11

ops/alpha-deploy.tmpl.yaml
afred: 👍

ops/provision/modules/networking/security-groups.tf
orangewolf: correct

ops/prod-deploy.tmpl.yaml
afred: 👍

ops/provision/modules/ec2/main.tf
afred: 👍

ops/provision/modules/k8s/files/prod_issuer.yaml
afred: 👍

ops/provision/modules/k8s/main.tf
afred: I found it under "Coordination > Leases" in the Rancher menu, but still wondering what it is used for

ops/provision/modules/networking/security-groups.tf
afred: 👍

ops/provision/modules/k8s/main.tf
afred: 👍 i guess i had a too-restrictive filter selected

ops/provision/modules/k8s/main.tf
afred: 👍

ops/provision/variables.tf
orangewolf: they have to exist in both places, but they dont actually need the default. I was being lazy

afred (approved):
👍

PR #684, opened 2022-11-01
upgrade hyrax chart to 1.6 including prerelease fix for service issue

orangewolf:
PR for Hyrax is here https://github.com/samvera/hyrax/pull/5907


afred (approved):
👍

PR #697, opened 2023-01-04
Sipity collision prevention

orangewolf:
Keep Sipity::Entity if it already exists. Do not update user in this case, to prevent hijacking of a record.


afred (approved):
👍

PR #706, opened 2023-01-30
admin data can be missing in imports, bring back alpha url

orangewolf:
- admin data can be missing on imported objects, this fixes that. 
- also brings back alpha url for viewing current prod w/ bulkrax turned on

### DIFF
--- a/app/models/bulkrax/pbcore_xml_entry.rb
+++ b/app/models/bulkrax/pbcore_xml_entry.rb
@@ -74,7 +74,7 @@ module Bulkrax
       xml_asset_id = self.raw_metadata['id']
       work = Asset.where(id: manifest_asset_id || xml_asset_id).first if manifest_asset_id || xml_asset_id
 
-      admin_data_gid =  if work.present?
+      admin_data_gid =  if work.present? && work.admin_data.present?
                           work.admin_data.update!(bulkrax_importer_id: bulkrax_importer_id)
                           work.admin_data_gid
                         else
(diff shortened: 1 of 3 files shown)


app/models/bulkrax/pbcore_xml_entry.rb
```diff
@@ -74,7 +74,7 @@ def update_or_create_admin_data_gid(bulkrax_importer_id)
       xml_asset_id = self.raw_metadata['id']
       work = Asset.where(id: manifest_asset_id || xml_asset_id).first if manifest_asset_id || xml_asset_id
 
-      admin_data_gid =  if work.present?
+      admin_data_gid =  if work.present? && work.admin_data.present?
```
afred, line 77: This is interesting because our business logic _should_ mean that all Assets that have been saved should also have a corresponding `AdminData` record, since creating/updating the `AdminData` record is in a `before_save` callback on the `Asset` model.

So if we are running into a situation here where `work.present?` is true, but `work.admin_data.present?` is false, that suggests a bad state of the data.

@orangewolf do you know under what conditions this ends up being the case?

orangewolf: @afred it happens when someone stupidly deletes a bunch of admin data records while trying to clean up database objects left over from the Picard maneuver work. As far as I could tell, the ones that got removed were all in their default state as they represented newly imported records.

afred (approved):
👍

PR #709, opened 2023-02-28
Bump ActiveFedora, Hyrax, Handle tombstones, fix admin_data on reimport

orangewolf:


app/jobs/bulkrax/import_work_job.rb
```diff
@@ -0,0 +1,44 @@
...
+      entry.save!
+      entry.importer.current_run = importer_run
+      entry.importer.record_status
+    rescue Bulkrax::CollectionsCreatedError, Ldp::NotFound => e
+      Rails.logger.warn("#{self.class} entry_id: #{entry_id}, run_id: #{run_id} encountered #{e.class}: #{e.message}")
+      # You get 3 attempts at the above perform before we have the import exception cascade into
+      # the Sidekiq retry ecosystem.
+      # rubocop:disable Style/IfUnlessModifier
+      if time_to_live <= 1
+        raise "Exhauted reschedule limit for #{self.class} entry_id: #{entry_id}, run_id: #{run_id}.  Attemping retries"
```
afred: "Exhausted" :)

afred: I assume that it's `entity.save!` that is throwing the `Ldp::NotFound` from which we are rescuing/retrying here. What is it that is not being found?

afred (dismissed):
👍

app/jobs/bulkrax/import_work_job.rb
orangewolf: Fedora is returning this periodically for POST requests. it doesn't appear to affect reads. I haven't seen it on any other repositories, only this one. I _think_ it happens with the garbage collection takes too long to finish and Fedora times out the request. It manifests by having 10-30 second periods where all posts 404. then it comes back and starts working like normal again. This seems to happen after we've been pulling records from the old id set.

app/jobs/bulkrax/import_work_job.rb
```diff
@@ -0,0 +1,44 @@
...
+      if time_to_live <= 1
+        raise "Exhausted reschedule limit for #{self.class} entry_id: #{entry_id}, run_id: #{run_id}.  Attemping retries"
+      end
+      # rubocop:enable Style/IfUnlessModifier
+      reschedule(entry_id, run_id, time_to_live)
+    end
+    # rubocop:enable Rails/SkipsModelValidations
+
+    def reschedule(entry_id, run_id, time_to_live)
+      ImportWorkJob.set(wait: 1.minute).perform_later(entry_id, run_id, time_to_live - 1)
```
afred, line 41: @orangewolf what are the conditions that lead to the `Ldp::NotFound`?

afred (approved):
👍

PR #710, opened 2023-03-06
deal with calls to /members and /list_source

orangewolf:
this should prevent assets from becoming unreachable.

### DIFF
--- a/config/initializers/hyrax.rb
+++ b/config/initializers/hyrax.rb
@@ -192,6 +192,8 @@ Hyrax.config do |config|
 
   ## Do not alter unless you understand how ActiveFedora handles URI/ID translation
   config.translate_id_to_uri = lambda do |id|
+    return "#{ActiveFedora.fedora.host}#{ActiveFedora.fedora.base_path}/#{::Noid::Rails.treeify(id)}" unless id.match(/^cpb-aacip/)
+    id, tail = id.split('/')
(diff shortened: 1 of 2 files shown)


config/initializers/hyrax.rb
```diff
@@ -192,6 +192,8 @@
 
   ## Do not alter unless you understand how ActiveFedora handles URI/ID translation
   config.translate_id_to_uri = lambda do |id|
+    return "#{ActiveFedora.fedora.host}#{ActiveFedora.fedora.base_path}/#{::Noid::Rails.treeify(id)}" unless id.match(/^cpb-aacip/)
+    id, tail = id.split('/')
```
afred, line 196: For ID's that will have a "tail", are these of the form "cpb-aacip/123-456789"?

orangewolf: the only two I've seen so far are
`cpb-aacip-7fdd93e156c/list_source` or `cpb-aacip-7fdd93e156c/members` but I wanted to cover the generic case so if there was something I wasn't seeing the logs we wouldn't have a sleeper on our hands

afred: Did some digging, and the ID's that come from the pbcore xml in the form `cpb-aacip/123-456789` are getting converted to `cpb-aacip-123-45678` in `PbcoreXMLMapper#normalize_aapb_id` and the `PBCoreXMLMapper` is being used by `PBCoreXMLParser` when initially parsing the record. Approving this

afred (approved):
👍

PR #712, opened 2023-04-11
Parser instantiation detaction bug

orangewolf:
Bulkrax was updating existing instantiations when it should have been making new ones. This is because local identifiers have no uniqueness guarantees. Instead we scope them with in the Asset now.

### DIFF
--- a/app/parsers/pbcore_xml_parser.rb
+++ b/app/parsers/pbcore_xml_parser.rb
@@ -149,7 +149,10 @@ class PbcoreXmlParser < Bulkrax::XmlParser
 
     pi_rows = pbcore_physical_instantiations.map do |inst|
       xml_pi = AAPB::BatchIngest::PBCoreXMLMapper.new(inst.to_xml).physical_instantiation_attributes
-      physical_instantiation = PhysicalInstantiation.where(local_instantiation_identifier: xml_pi[:local_instantiation_identifier]).first&.attributes&.symbolize_keys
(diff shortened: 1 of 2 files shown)


app/parsers/pbcore_xml_parser.rb
```diff
@@ -149,7 +149,10 @@ def set_objects(file, index)
 
     pi_rows = pbcore_physical_instantiations.map do |inst|
       xml_pi = AAPB::BatchIngest::PBCoreXMLMapper.new(inst.to_xml).physical_instantiation_attributes
-      physical_instantiation = PhysicalInstantiation.where(local_instantiation_identifier: xml_pi[:local_instantiation_identifier]).first&.attributes&.symbolize_keys
+      # Find members of the asset that have the same class and local identifier. If no asset, then no physical instantiation can exist
+      physical_instantiation = asset.members.detect do |member|
+          member.local_instantiation_identifier == xml_pi[:local_instantiation_identifier] && member.class == PhysicalInstantiation
+        end if asset && xml_pi[:local_instantiation_identifier].present?
```
afred, line 155: ☝️ unindent 2 spaces i think, yeah?

afred (approved):
👍

app/parsers/pbcore_xml_parser.rb
orangewolf: yep. I'll merge and then handle so we don't have to re-review

PR #713, opened 2023-04-11
Double include

orangewolf:
only include the basic metadata once per model. no real bug we are fixing here, just cleaning up something that could be a problem at some point.


app/models/digital_instantiation.rb
```diff
@@ -159,7 +159,4 @@ def save_instantiation_admin_data
     find_or_create_instantiation_admin_data
     self.instantiation_admin_data.save
   end
-
-  # This must be included at the end, because it finalizes the metadata if you have any further properties define above in current model
-  include ::Hyrax::BasicMetadata
```
afred, line 164: @orangewolf is the comment here no longer a concern?

afred (approved):
👍 approved on condition that the comment accompanying the extra include is no longer a concern.

app/models/digital_instantiation.rb
orangewolf: the one above says the same thing - it has to be after any properties. putting it under the private is the issue

PR #714, opened 2023-04-11
Jemalloc

orangewolf:
Big memory management improvements across the board, but especially in multithreaded puma and sidekiq.See https://medium.com/motive-eng/we-solved-our-rails-memory-leaks-with-jemalloc-5c3711326456 and https://klimo.io/posts/case-study-double-your-throughput-jemalloc-heroku/ for background info. The biggest thing was we used `MALLOC_ARENA_MAX` to prevent memory growth over time before in Hyrax. But `MALLOC_ARENA_MAX` was causing massive performance hit.


afred (approved):
👍 changes to Dockerfile make sense, but haven't tried to build locally. Is that something you think i should try?

Also, looks like this is based of another branch with a current outstanding PR

orangewolf:
> 👍 changes to Dockerfile make sense, but haven't tried to build locally. Is that something you think i should try?
> 
> Also, looks like this is based of another branch with a current outstanding PR

re-running CI which will build the image for us. if that's green I'll merge. Build failed due to rubygems being down briefly yesterday.

PR #716, opened 2023-04-25
Refactor Bulkrax Identifer to use GUIDs, Fix uniqueness of Instantiations and Essence Tracks

orangewolf:
also resolves the Instantiation show page when Asset is not connected yet.

Refs:
https://github.com/scientist-softserv/dev-ops/issues/656
https://github.com/scientist-softserv/dev-ops/issues/660
https://github.com/scientist-softserv/dev-ops/issues/661

This likely breaks CSV and Manifest imports. It also likely breaks the specs. We're going to fix those as a separate PR as I'm feeling every second we are not importing


config/initializers/hyrax.rb
```diff
@@ -191,9 +191,17 @@
   # config.lock_time_to_live = 60_000
 
   ## Do not alter unless you understand how ActiveFedora handles URI/ID translation
+  # config.translate_id_to_uri = lambda do |id|
+  #   uri = "#{ActiveFedora.fedora.host}#{ActiveFedora.fedora.base_path}/#{::Noid::Rails.treeify(id)}"
+  #   Rails.logger.error("============ #{uri} ============")
+  #   return uri
+  # end
```
alishaevn: did you mean to leave this?

config/initializers/bulkrax.rb
```diff
@@ -32,7 +32,7 @@ def headers(term = nil)
       properties
     end
 
-    config.fill_in_blank_source_identifiers = ->(obj, index) { "#{obj.importerexporter.id}-#{index}" }
+    config.fill_in_blank_source_identifiers = ->(obj, type, index, key_count) { "#{type}-#{index}-#{key_count}" }
```
alishaevn: is there a reason that `obj` was left as an argument although we aren't using it anymore?

app/parsers/pbcore_manifest_parser.rb
```diff
@@ -147,7 +147,7 @@ def set_objects
       asset_bulkrax_identifier =  if asset.bulkrax_identifier
                                     asset.bulkrax_identifier
                                   else
-                                    Bulkrax.fill_in_blank_source_identifiers.call(self, "Asset-#{index}-1")
+                                    Bulkrax.fill_in_blank_source_identifiers.call(self, "Asset", asset_id, 1)
```
alishaevn:
```suggestion
                                    Bulkrax.fill_in_blank_source_identifiers.call(self, "Asset", index, 1)
```

app/parsers/pbcore_xml_parser.rb
```diff
@@ -74,55 +69,56 @@ def create_works
...
+      set_objects(file, index).each do |record|
+        break if limit_reached?(limit, index)
+        seen[record[work_identifier]] = true
+      end
+      increment_counters(index)
+    end
+  rescue StandardError => e
+    status_info(e)
+  end
+
```
alishaevn:
```suggestion
```

alishaevn: I didn't see either of these methods being called in the changed code. were they left over from our mobbing session? or did I overlook something?

alishaevn (changes requested):

config/initializers/hyrax.rb
orangewolf: I did, I'm adding a comment as to why - its the original version and we want to put it back once we finish moving old records out.

config/initializers/bulkrax.rb
orangewolf: good catch - this wasn't even consistant across the importers, some had 3 args, some had 4. fixed it

app/parsers/pbcore_manifest_parser.rb
orangewolf: we explicitly want asset_id here - we're done with marking these 1,2,3. we want the pbcore GUID instead so they can easily be traced back to the pbcore data. by using this externally consistant key, we can always find the entries and all items connected to a specific xml file.

alishaevn: ok. I thought we were going with something more representative like the `id`, but each of the other calls to this method still use `index` as the value being passed in.

app/parsers/pbcore_xml_parser.rb
orangewolf: these are super useful for debugging and fixing things. I need one of them for the coming bulkrax entry migration script. I've added comments trying to justify their existence. Does that seem like enough to keep them around?

app/parsers/pbcore_xml_parser.rb
```diff
@@ -74,55 +69,62 @@ def create_works
     status_info(e)
   end
 
+  ##
+  # This method is usefuly for updating existing entries with out reimporting the works themselves
```
alishaevn:
```suggestion
  # This method is useful for updating existing entries with out reimporting the works themselves
```

alishaevn (changes requested):

app/parsers/pbcore_xml_parser.rb
alishaevn: yes! that makes sense.

alishaevn (dismissed):

app/parsers/pbcore_manifest_parser.rb
orangewolf: the other calls were existing methods - I changed what was passed in to index in the method, but I can be more consistent about it. Let me push up more changes. I am leaving the actual lambda alone - since it shouldn't care what value is passed in there keeping it called `index` makes sense to me, even if we only send it the GUID for now. it has a separation of concerns.

alishaevn: thanks for the clarification and explicit argument names!

alishaevn (approved):

.env
```diff
@@ -10,7 +10,7 @@ MYSQL_ROOT_PASSWORD=DatabaseFTW
 MYSQL_USER=root
 REDIS_HOST=redis
 REDIS_SERVER=redis
-SETTINGS__BULKRAX__ENABLED=false
+SETTINGS__BULKRAX__ENABLED=true
```
afred, line 13: @orangewolf is this going out to production ☝️

afred (approved):
👍 just one comment on the BULKRAX_ENABLED setting and whether it will end up on production or not.

.env
orangewolf: just for the record here - no, it will only be in the dev environment

PR #722, opened 2023-05-11
Spec fixes

bkiahstroud:


app/actors/hyrax/actors/asset_actor.rb
```diff
@@ -117,7 +117,9 @@ def annotation_attributes
         end
 
         def find_or_create_admin_data(env)
-          return env.curation_concern.admin_data if env.curation_concern.admin_data.present?
+          # Don't use present? here. It is broken for some ActiveRecord objects. Maybe due to overriding getter
+          # See: https://stackoverflow.com/questions/35410668/what-determines-the-return-value-of-present
+          return env.curation_concern.admin_data if env.curation_concern.admin_data.is_a?(AdminData)
```
afred, line 122: @bkiahstroud was `present?` returning false positive or false negative here? And do you know what the value of `env.curation_concern.admin_data` was in that circumstance?

I ask because from a business logic perspective, `env.curation_concern.admin_data` should never be anything other than `AdminData` instance or `nil`, so if it's winding up in a 3rd state, that might be an error state we need to look into preventing.

spec/factories/digital_instantiation.rb
```diff
@@ -8,7 +8,7 @@
     duration { '1:23:45' }
     file_size { '12354435234' }
     local_instantiation_identifier { ["1234"] }
-    digital_instantiation_pbcore_xml { Rack::Test::UploadedFile.new(Rails.root.join('spec/fixtures/sample_instantiation_valid.xml'), 'text/xml') }
+    digital_instantiation_pbcore_xml { File.open(Rails.root.join('spec/fixtures/sample_instantiation_valid.xml')) }
```
afred, line 11: The `digital_instantiation_pbcore_xml` is defined in `DigitalInstantiation` class with
```
mount_uploader :digital_instantiation_pbcore_xml, PbCoreInstantiationXmlUploader
```
I don't recall offhand which kind of object that this ends up being, but there 3 places where the code expects the object to have some method on it:
* `digital_instantiation_pbcore_xml.file`
* `digital_instantiation_pbcore_xml.file.file`
* `digital_instantiation_pbcore_xml.file.tempfile`

I think `File.open(...)` would have to return a similar object in order to work but not sure if it does.

app/actors/hyrax/actors/asset_actor.rb
bkiahstroud: It was returning a false negative. `env.curation_concern.admin_data` was returning an instance of `AdminData`. 

```ruby
env.curation_concern.admin_data
=> #<AdminData id: ...
env.curation_concern.admin_data.present?
=> false
```

spec/factories/digital_instantiation.rb
bkiahstroud: I just tried these three methods using a debugger locally: 

- `File.open` 
- `Rack::Test::UploadedFile.new` 
- `fixture_file_upload` 

All three result in `digital_instantiation_pbcore_xml` being an instance of `PbCoreInstantiationXmlUploader` with, as far as I can tell, identical behavior. 

@sephirothkod may have more info on this. I don't know much beyond "this wasn't working before, this change seems to make it work better"

bkiahstroud:
Closing in favor of https://github.com/WGBH-MLA/ams/pull/723, which should allow the CI to run

spec/factories/digital_instantiation.rb
sephirothkod: The `Rack::Test::UploadedFile.new` call for some reason was passing through an empty (0 byte) file and causing a large number of tests to fail. I am unsure as to why this was happening but as soon as I switched to using `File.new` it was able to pull in the entire file and the tests passed.

PR #723, opened 2023-05-12
Spec fixes

bkiahstroud:


app/actors/hyrax/actors/asset_actor.rb
```diff
@@ -117,7 +117,9 @@ def annotation_attributes
         end
 
         def find_or_create_admin_data(env)
-          return env.curation_concern.admin_data if env.curation_concern.admin_data.present?
+          # Don't use present? here. It is broken for some ActiveRecord objects. Maybe due to overriding getter
+          # See: https://stackoverflow.com/questions/35410668/what-determines-the-return-value-of-present
+          return env.curation_concern.admin_data if env.curation_concern.admin_data.is_a?(AdminData)
```
bkiahstroud, line 122: Discussion brought over from [the original PR](https://github.com/WGBH-MLA/ams/pull/722): 

> @bkiahstroud was `present?` returning false positive or false negative here? And do you know what the value of `env.curation_concern.admin_data` was in that circumstance?
>
> I ask because from a business logic perspective, `env.curation_concern.admin_data` should never be anything other than `AdminData` instance or `nil`, so if it's winding up in a 3rd state, that might be an error state we need to look into preventing.
> \- @afred 
> > It was returning a false negative. `env.curation_concern.admin_data` was returning an instance of `AdminData`.
> >
> > ```ruby
> > env.curation_concern.admin_data
> > => #<AdminData id: ...
> > env.curation_concern.admin_data.present?
> > => false
> >```
> > \- @bkiahstroud

spec/factories/digital_instantiation.rb
```diff
@@ -8,7 +8,7 @@
     duration { '1:23:45' }
     file_size { '12354435234' }
     local_instantiation_identifier { ["1234"] }
-    digital_instantiation_pbcore_xml { Rack::Test::UploadedFile.new(Rails.root.join('spec/fixtures/sample_instantiation_valid.xml'), 'text/xml') }
+    digital_instantiation_pbcore_xml { File.open(Rails.root.join('spec/fixtures/sample_instantiation_valid.xml')) }
```
bkiahstroud, line 11: Discussion brought over from [the original PR](https://github.com/WGBH-MLA/ams/pull/722): 

> The `digital_instantiation_pbcore_xml` is defined in `DigitalInstantiation` class with
>
> ```
> mount_uploader :digital_instantiation_pbcore_xml, PbCoreInstantiationXmlUploader
> ```
> I don't recall offhand which kind of object that this ends up being, but there 3 places where the code expects the object to have some method on it:
>
> - `digital_instantiation_pbcore_xml.file`
> - `digital_instantiation_pbcore_xml.file.file`
> - `digital_instantiation_pbcore_xml.file.tempfile`
>
> I think `File.open(...)` would have to return a similar object in order to work but not sure if it does.
> \- @afred 
> > I just tried these three methods using a debugger locally:
> >
> > - `File.open`
> > - `Rack::Test::UploadedFile.new`
> > - `fixture_file_upload`
> >
> > All three result in `digital_instantiation_pbcore_xml` being an instance of `PbCoreInstantiationXmlUploader` with, as far as I can tell, identical behavior.
> >
> > @sephirothkod may have more info on this. I don't know much beyond "this wasn't working before, this change seems to make it work better"
> > \- @bkiahstroud 
> > > The `Rack::Test::UploadedFile.new` call for some reason was passing through an empty (0 byte) file and causing a large number of tests to fail. I am unsure as to why this was happening but as soon as I switched to using `File.new` it was able to pull in the entire file and the tests passed.
> > > \- @sephirothkod

afred: 👍

app/actors/hyrax/actors/asset_actor.rb
afred: @bkiahstroud i went digging for why this happens because it seems like a real code smell to me to have `.present?` return `false` (and `.blank?` and `.empty?` return `true`), when in fact the `AdminData` instance is a persisted record in the database.

I found [this PR](https://github.com/WGBH-MLA/ams/pull/102/files), which adds the mix-in `EmptyDetection`, which overrides `empty?` method to return true when all `AdminData` fields (except for `id`, `inspection_id`, `created_at`, `updated_at`) are either nil, an empty array, or an array with a single element that is an empty string.

As far as I can tell, this weird definition of `empty?` is used in [`asset_form.rb`](https://github.com/WGBH-MLA/ams/blob/develop/app/forms/hyrax/asset_form.rb#L82), [`digital_instantiation_form.rb`](https://github.com/WGBH-MLA/ams/blob/ef53626743aaf437e9f35cb684797003dc37c766/app/forms/hyrax/digital_instantiation_form.rb#L48), and [`physical_instantiation_form.rb`](https://github.com/WGBH-MLA/ams/blob/ef53626743aaf437e9f35cb684797003dc37c766/app/forms/hyrax/physical_instantiation_form.rb#L53)

IMO, this is pretty ugly, and _really_ violates the principle of least surprise, given how it affects other common methods like `.blank?` and `.present?`. I approved the PR this came in on, however, I failed to flag this one.

Your change here makes sense given the `EmptyDetection` weirdness, but it's probably a good idea to stop overriding `.empty?` for `AdminData` and `InstantiationAdminData` and have a different method used for detecting that specific kind of emptiness.

afred (approved):
👍

PR #725, opened 2023-06-06
Migrate OLD GUIDs and Fix the Specs

orangewolf:
this fixes specs and should resolve a "admindata cant be updated" when admindata is blank bug. It also adds a rake task for migrating GUIDs in the fcrepo tree.


afred (approved):
👍

PR #728, opened 2023-07-14
Prevent Assets missing children from being pushed to AAPB (Validator)

bkiahstroud:
# Summary

Adds new validations the ensures an `Asset` has all of it's children (and that it's children have their children) present in the database and associated with it before pushing it to AAPB.

### Demo: https://share.getcloudapp.com/P8uX1kN0

image

### DIFF
--- a/app/models/push.rb
+++ b/app/models/push.rb
@@ -6,6 +6,7 @@ class Push < ApplicationRecord
     errors.add(:user, "is required") unless user.is_a? User
     missing_ids = ids_not_found
     errors.add(:pushed_id_csv, "The following IDs are not found in the repository: #{missing_ids.join(', ')}") unless missing_ids.empty?
+    validate_missing_children
   end
 
   # NOTE: do not memoize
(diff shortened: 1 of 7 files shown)


app/models/push.rb
```diff
@@ -27,6 +28,24 @@ def ids_not_found
...
+      return if invalid_docs.blank?
+
+      add_status_error(invalid_docs, Asset::VALIDATION_STATUSES[:missing_children])
+    end
+
+    def add_status_error(invalid_docs, status)
+      ids_matching_status = invalid_docs.select { |doc| doc.validation_status_for_aapb.include?(status) }.map(&:id)
+      # Prevents adding errors to docs that don't have a value
+      # in :validation_status_for_aapb, including all non-Assets.
+      return if ids_matching_status.blank?
```
bkiahstroud, line 44: This means non-Asset records won't show an error (I _think_ only Assets get pushed anyways, so I'm not sure if this is even relevant).

More importantly, it means historic Assets that haven't been updated since this change won't throw errors. 

Is this what we want?

orangewolf (approved):

PR #730, opened 2023-07-18
i729 Adjust logic determining if an Asset is missing children

bkiahstroud:
Using "less than" instead of "does not equal" fixes a couple issues:

- If a new child record was created through the UI, the Asset would be marked as missing children
- If a pre-existing Asset, whose :intended_children_count had not been set, was updated outside of Bulkrax, it would be marked as missing children

### DIFF
--- a/app/actors/hyrax/actors/asset_actor.rb
+++ b/app/actors/hyrax/actors/asset_actor.rb
@@ -228,7 +228,7 @@ module Hyrax
           intended_children_count = env.curation_concern.intended_children_count.to_i
           current_children_count = env.curation_concern.all_members.size
 
-          if current_children_count != intended_children_count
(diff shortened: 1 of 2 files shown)


app/actors/hyrax/actors/asset_actor.rb
```diff
@@ -228,7 +228,7 @@ def set_validation_status(env)
           intended_children_count = env.curation_concern.intended_children_count.to_i
           current_children_count = env.curation_concern.all_members.size
 
-          if current_children_count != intended_children_count
```
jeremyf, line 231: Good catch!

jeremyf (approved):

alishaevn (approved):

PR #732, opened 2023-07-24
i10 Script to set number of intended children for all Assets

bkiahstroud:
Ref 
- https://github.com/scientist-softserv/ams/issues/10

### DIFF
--- a/app/actors/hyrax/actors/asset_actor.rb
+++ b/app/actors/hyrax/actors/asset_actor.rb
@@ -53,19 +53,19 @@ module Hyrax
         def set_admin_data_attributes(admin_data, env)
           AdminData.attributes_for_update.each do |k|
             # Some attributes are serialized on AdminData, so always send an array
-            if should_empty_admin_data_value?(k, admin_data, env)
-              AdminData::SERIALIZED_FIELDS.include?(k) ? admin_data.send("#{k}=", Array.new) : admin_data.send("#{k}=", "")
+            if should_empty_admin_data_value?(k, env)
+              AdminData::SERIALIZED_FIELDS.include?(k) ? admin_data.send("#{k}=", Array.new) : admin_data.send("#{k}=", nil)
             elsif env.attributes[k].present?
               AdminData::SERIALIZED_FIELDS.include?(k) ? admin_data.send("#{k}=", Array(env.attributes[k])) : admin_data.send("#{k}=", env.attributes[k].to_s)
             end
           end
         end
 
-        def should_empty_admin_data_value?(key, admin_data, env)
-          key != :bulkrax_importer_id &&
-          env.attributes.key?(key) &&
-          admin_data.send(key).present? &&
-          !env.attributes[key].present?
+        def should_empty_admin_data_value?(key, env)
+          return false if %i[bulkrax_importer_id hyrax_batch_ingest_batch_id].include?(key)
+
+          # The presence of the key with a "blank" value indicates we're intentionally emptying the value
+          env.attributes.key?(key) && env.attributes[key].blank?
         end
 
         def delete_removed_annotations(admin_data, env)
--- a/app/actors/hyrax/actors/asset_actor.rb
+++ b/app/actors/hyrax/actors/asset_actor.rb
@@ -63,6 +63,7 @@ module Hyrax
 
         def should_empty_admin_data_value?(key, admin_data, env)
           key != :bulkrax_importer_id &&
+          env.attributes.key?(key) &&
           admin_data.send(key).present? &&
           !env.attributes[key].present?
         end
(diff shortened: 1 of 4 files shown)


app/services/ams/backfill_asset_validation_status.rb
```diff
@@ -0,0 +1,164 @@
...
+                        else
+                          raise StandardError, "Unable to find source data for Asset #{asset.id}"
+                        end
+
+      parsed_source_data = AAPB::BatchIngest::PBCoreXMLMapper.new(raw_source_data).asset_attributes
+      attrs_for_actor['intended_children_count'] = parsed_source_data[:intended_children_count]
+      if attrs_for_actor['intended_children_count'].blank?
+        raise StandardError, "Unable to count intended children for Asset #{asset.id}"
+      end
+
```
orangewolf, line 115: if you havent loaded the asset here - you can create a background job to do the actor stack and save and never load the asset in this loop.

orangewolf: sidebar: what if there was an activerecord object called "CleanupRun" that had id, item_id, item_type, job_name, status, error columns.

app/services/ams/backfill_asset_validation_status.rb
```diff
@@ -0,0 +1,180 @@
...
+    ALL_IDS_PATH = WORKING_DIR.join('all_ids.txt').freeze
+    PROCESSED_IDS_PATH = WORKING_DIR.join('processed_ids.txt').freeze
+    REMAINING_IDS_PATH = WORKING_DIR.join('remaining_ids.txt').freeze
+    FAILED_IDS_PATH = WORKING_DIR.join('failed_ids.txt').freeze
+
+    attr_accessor :ids, :logger
+
+    def initialize
+      setup_working_directory
+      # TODO: replace with tagged logger
```
bkiahstroud, line 17: Non-blocking

app/actors/hyrax/actors/asset_actor.rb
```diff
@@ -63,6 +63,7 @@ def set_admin_data_attributes(admin_data, env)
 
         def should_empty_admin_data_value?(key, admin_data, env)
           key != :bulkrax_importer_id &&
+          env.attributes.key?(key) &&
```
jeremyf: This is a bit confusing.  And thinking about the how we can short circuit calling admin data; I suspect that's slower.

```ruby
key != :bulkrax_importer_id &&
env.attributes.key?(key) &&
!env.attributes[key].present? &&
admin_data.send(key).present?
```

jeremyf (approved):

app/actors/hyrax/actors/asset_actor.rb
bkiahstroud: > And thinking about the how we can short circuit calling admin data; I suspect that's slower.

@jeremyf can you explain what you mean here? What is slower?

jeremyf: Sure.  `env.attributes.key?` is almost certainly a hash access.  Whereas `admin_data` is an object and we're sending it the `key` message.  What does that method do?  It could query all of Fedora.

Also, by having that `admin_data` sandwiched between two `env.attributes` methods there's a bit of a disjointed logical flow; namely, I'm going from one subject to another and then back to that first subject.  And with the `&&` that feels a bit much.

app/actors/hyrax/actors/asset_actor.rb
```diff
@@ -53,19 +53,19 @@ def save_aapb_admin_data(env)
...
-        def should_empty_admin_data_value?(key, admin_data, env)
-          key != :bulkrax_importer_id &&
-          env.attributes.key?(key) &&
-          admin_data.send(key).present? &&
-          !env.attributes[key].present?
+        def should_empty_admin_data_value?(key, env)
+          return false if %i[bulkrax_importer_id hyrax_batch_ingest_batch_id].include?(key)
+
+          # The presence of the key with a "blank" value indicates we're intentionally emptying the value
+          env.attributes.key?(key) && env.attributes[key].blank?
```
jeremyf, line 68: What happened to the `admin_data` check?

bkiahstroud: It seemed unnecessary. It was saying: 

> Only null out the value if a value is present 

I see no danger in the following, so I removed it: 

```ruby
admin_data.my_attribute
=> nil

admin_data.my_attribute = nil

admin_data.my_attribute
=> nil
```

laritakr (approved):

app/actors/hyrax/actors/asset_actor.rb
jeremyf: Agreed

PR #737, opened 2023-08-07
🛠️ Reorder migrations

kirkkwang:
There was a problem booting this application up where it complained about dropping a foreign key that didn't exist.  Here's the error:

```sh
Mysql2::Error: Cannot drop index 'index_bulkrax_entries_on_importer_id':
needed in a foreign key constraint: ALTER TABLE `bulkrax_entries` CHANGE
`importer_id` `importerexporter_id` bigint(20) DEFAULT NULL
```

This commit makes it so `RemoveForeignKeyFromBulkraxEntries` runs earlier to get around this problem.


kirkkwang:
favoring 
- #738

PR #739, opened 2023-08-07
🤖 Remove `retag` step

kirkkwang:
This commit will remove the `retag` step from GitHub Actions.  This was not working as intended and was causing the CI to be red.

Ref:
- https://github.com/scientist-softserv/ams/issues/44


ShanaLMoore (approved):
remove per Rob.

PR #746, opened 2023-08-09
Block pushing assets to AAPB until their validation status is set

laritakr:
- Set Asset.validation_status_for_aapb as part of asset_actor
- Do not allow asset to be pushable unless validation_status_for_aapb is valid.

Ref https://github.com/scientist-softserv/ams/issues/51

<details>
<summary> For non-asset id </summary>

Screenshot 2023-08-09 at 6 10 07 PM
</details>
<details>
<summary> For asset id not yet validated </summary>

Screenshot 2023-08-09 at 6 20 00 PM
</details>
<details>
<summary> For asset id with empty validation status</summary>

Screenshot 2023-08-09 at 6 38 06 PM
</details>

### DIFF
--- a/app/actors/hyrax/actors/asset_actor.rb
+++ b/app/actors/hyrax/actors/asset_actor.rb
@@ -235,7 +235,9 @@ module Hyrax
           current_children_count = env.curation_concern.all_members.reject { |child| child.is_a?(Contribution) }.size
           intended_children_count = env.curation_concern.intended_children_count.to_i
 
-          if current_children_count < intended_children_count
+          if env.curation_concern.intended_children_count.nil? && env.curation_concern.validation_status_for_aapb.empty?
+            env.curation_concern.validation_status_for_aapb = [Asset::VALIDATION_STATUSES[:status_not_validated]]
+          elsif current_children_count < intended_children_count
             env.curation_concern.validation_status_for_aapb = [Asset::VALIDATION_STATUSES[:missing_children]]
           else
             env.curation_concern.validation_status_for_aapb = [Asset::VALIDATION_STATUSES[:valid]]
--- a/app/actors/hyrax/actors/asset_actor.rb
+++ b/app/actors/hyrax/actors/asset_actor.rb
@@ -235,7 +235,7 @@ module Hyrax
           current_children_count = env.curation_concern.all_members.reject { |child| child.is_a?(Contribution) }.size
           intended_children_count = env.curation_concern.intended_children_count.to_i
 
-          if env.curation_concern.intended_children_count.nil? && env.curation_concern.validation_status_for_aapb.empty?
+          if env.curation_concern.intended_children_count.blank? && env.curation_concern.validation_status_for_aapb.blank?
             env.curation_concern.validation_status_for_aapb = [Asset::VALIDATION_STATUSES[:status_not_validated]]
           elsif current_children_count < intended_children_count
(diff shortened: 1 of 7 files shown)


orangewolf:
I agree I'd like Kiah's eyes on this one. also some spec work to do.

ShanaLMoore:
@orangewolf we may need your help with running specs locally. Are you able to without timeout or selenium issues? LaRita mentioned she was having issues with this earlier today. She can run an individual spec OK but not the entire suite at once without such errors. 

Since we pushed it to the ci she'll know which ones to address individually, so that may be an ok workaround for now.

app/actors/hyrax/actors/asset_actor.rb
```diff
@@ -235,7 +235,9 @@ def set_validation_status(env)
           current_children_count = env.curation_concern.all_members.reject { |child| child.is_a?(Contribution) }.size
           intended_children_count = env.curation_concern.intended_children_count.to_i
 
-          if current_children_count < intended_children_count
+          if env.curation_concern.intended_children_count.nil? && env.curation_concern.validation_status_for_aapb.empty?
```
bkiahstroud:
```suggestion
          if env.curation_concern.intended_children_count.nil? && env.curation_concern.validation_status_for_aapb.blank?
```

I _think_ `#empty?` here might throw an error. I think there's a chance records created before the `:validation_status_for_aapb` property was introduced will return `nil` instead of `[]`. 

Even if that's not the case, I tend to almost always use `#blank?` since it covers more cases: 

```ruby
irb(main):001:0> nil.empty?
Traceback (most recent call last):
        1: from (irb):1
NoMethodError (undefined method `empty?' for nil:NilClass)

irb(main):002:0> nil.blank?
=> true
```

I'm not really worried about `intended_children_count.nil?` since `:intended_children_count` is singular and thus should be `nil` if it's unset (though I'd personally probably still use `#blank?` 😅)

bkiahstroud (changes requested):

app/models/push.rb
```diff
@@ -34,11 +34,18 @@ def validate_status
...
+        add_status_error(invalid_docs, status_value)
+      end
     end
 
     def add_status_error(invalid_docs, status)
-      ids_matching_status = invalid_docs.select { |doc| doc.validation_status_for_aapb.include?(status) }.map(&:id)
+      if status == Asset::VALIDATION_STATUSES[:empty]
+        ids_matching_status = invalid_docs.select { |doc| doc.validation_status_for_aapb.blank? }.map(&:id)
+      else
+       ids_matching_status = invalid_docs.select { |doc| doc.validation_status_for_aapb.include?(status) }.map(&:id)
```
bkiahstroud: 1. Single space indentation instead of double 
2. Is this variable visible further down in the method? I assume so based on the screenshots, but I thought declaring a variable in a block meant it wasn't in scope outside of the block 🤔

app/actors/hyrax/actors/asset_actor.rb
laritakr: updated

app/models/push.rb
laritakr: It does seem strange that it is visible later but it definitely is. Regardless, to be more clear, I will initialize it outside of the if block.

bkiahstroud (approved):
Looks great! :shipit:

PR #753, opened 2023-08-12
I60 config discrepancies destroys development data when running specs

ShanaLMoore:
In this PR, I compared configurations against working apps like hyku proper and UTK, and took note of the configuration discrepancies. 

In particular, we were trying to find the reason why running specs would destroy our local data. Specs should rely on its test environment and its data should be separate from development. Using RAILS_ENV=test didn't make a difference.

For specs there's helper method called #reset! It runs in a before suite hook. Within that method, ActiveFedora::Cleaner.clean! triggered the issue of destroying development data. 

With LaRitas help we discovered the issue was actually because of solr using hydra-development for both environments.

Issue: 
- https://github.com/scientist-softserv/ams/issues/60#issuecomment-1675496108

### EOD TODO: 

- test out changes to confirm that running specs won't cause development data to be destroyed

ref: https://github.com/scientist-softserv/ams/issues/60#issuecomment-1675496108

1. Assuming you have no data, create data in rails console `3.times {Asset.create!}`
2. Verify count `Asset.count` => 3
3. Run the following spec
```ruby
DEPENDENCIES_TEST=1 bundle exec rspec spec/features/update_admin_data_spec.rb 
```
4. `reload!` ; `Asset.count` => hopefully returns the correct count (ie: 3 in this example)

### UPDATE: Tested ✅ 
image

In this screenshot, i created 3 assets in rails console. then in the spec context i ran the offensive line that previously deleted development data, reloaded rails c and asked for the count again. Previously it would return 0 but it appears to be persisting now.


ShanaLMoore:
@orangewolf @aprilrieger I would appreciate a review of this, from a devops perspective. I'm not confident about or totally familiar with the consequences of my changes besides the fact that its no longer destroying my development data when I run a spec 🎉 

Also, something weird is still happening. I am able to connect to hyku.test even though it's not listed in the docker-compose file and i should only be using ams.test. [This](https://github.com/WGBH-MLA/ams/commit/8c69a372033b5495527e5e9515eef1e0b0eb53ea) was also the reason why we weren't able to run feature specs locally.

ShanaLMoore:
Switching to Draft. It's ready for review but I don't want anyone to accidentally merge it until devops review.

.solr_wrapper
```diff
@@ -6,3 +6,9 @@ collection:
  persist: true
  dir: solr/config/
  name: hydra-development
+
```
ShanaLMoore: revert - we aren't using solr_wrapper

config/solr_wrapper_test.yml
```diff
@@ -8,3 +8,9 @@ collection:
     persist: true
     dir: solr/config
     name: hydra-test
+
```
ShanaLMoore: revert

docker-compose.yml
```diff
@@ -83,6 +83,13 @@ services:
       - OOM=script
       - VIRTUAL_PORT=8983
       - VIRTUAL_HOST=solr.ams.test
+      - SOLR_ADMIN_USERNAME=solr
```
ShanaLMoore: Don't need

orangewolf: these dont do anything

.solr_wrapper
orangewolf: revert since we dont use this

config/fedora.yml
```diff
@@ -11,6 +11,6 @@ test:
 production:
   user: fedoraAdmin
   password: fedoraAdmin
-  url: <%= ENV['FCREPO_URL'] %>
-  base_path: <%= ENV['FCREPO_BASE_PATH'] %>
+  url: <%= ENV['FCREPO_URL'] || "http://127.0.0.1:8983/fedora/rest" %>
```
orangewolf: un-needed change

config/solr_wrapper_test.yml
orangewolf: revert since we cant test it

orangewolf (changes requested):
remove stuff unrelated to this fix. the fix looks great though.

orangewolf (approved):

PR #755, opened 2023-08-14
I24 dual boot support for docker && ci

ShanaLMoore:
WIP

ref: https://github.com/scientist-softserv/ams/issues/24

resources: 
1. https://www.fastruby.io/blog/upgrade-rails/dual-boot/dual-boot-with-rails-6-0-beta.html
2. https://blog.testdouble.com/posts/2023-01-16-bootboot-in-rails-4.x-deployment-environments-with-docker/
3. https://github.com/Shopify/bootboot/issues/26

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -83,13 +83,6 @@ services:
       - OOM=script
       - VIRTUAL_PORT=8983
       - VIRTUAL_HOST=solr.ams.test
-      - SOLR_ADMIN_USERNAME=solr
-      - SOLR_ADMIN_PASSWORD=SolrRocks
-      - SOLR_COLLECTION=hydra-development
-      - SOLR_CLOUD_BOOTSTRAP=yes
-      - SOLR_ENABLE_CLOUD_MODE=yes
-      - SOLR_ENABLE_AUTHENTICATION=yes
-      - ZK_HOST=zoo:2181
     user: root
     command: bash -c "
       chown -R 8983:8983 /var/solr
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -83,6 +83,13 @@ services:
       - OOM=script
       - VIRTUAL_PORT=8983
       - VIRTUAL_HOST=solr.ams.test
+      - SOLR_ADMIN_USERNAME=solr
+      - SOLR_ADMIN_PASSWORD=SolrRocks
+      - SOLR_COLLECTION=hydra-development
+      - SOLR_CLOUD_BOOTSTRAP=yes
+      - SOLR_ENABLE_CLOUD_MODE=yes
+      - SOLR_ENABLE_AUTHENTICATION=yes
```
orangewolf: these are unneeded

orangewolf (changes requested):

orangewolf (approved):
excellent work!

ShanaLMoore:
@orangewolf we just noticed that something is still not right with this. 

We expect 10 failures to be reported from the rails 6 context.

We can see that rails 6 isn't using the correct gem versions. For example, hyrda-access-controls should be 12.0.1 as seen when running the rails 6 specs locally (and as stated in the Gemfile_next.lock file). 

image (24)

However, all the specs are oddly passing in the rails 6 build and it's not using the correct gem versions: 

image

I will open another PR to hopefully fix this.

PR #761, opened 2023-08-16
Deployed rails 6

orangewolf:
demo and bravo should be updated to rails 6. Demo is its own seperate db. Bravo points at the same db as prod does.

ref: https://github.com/scientist-softserv/ams/issues/71

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -30,13 +30,7 @@ jobs:
     container: dtzar/helm-kubectl:3.9.4
     environment: ${{ inputs.environment }}
     env:
-      HELM_EXPERIMENTAL_OCI: 1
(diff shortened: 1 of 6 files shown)


ShanaLMoore:
@orangewolf Please review https://github.com/WGBH-MLA/ams/pull/755#issuecomment-1680731291. It might apply to your work for this ticket.

ShanaLMoore (changes requested):
Thanks for doing this Rob. So they'll be testing rails 6 on demo and bravo? Can we deploy this branch to test it out? 

This PR is approved but there is a merge conflict. I would also like your input on this [issue](https://github.com/WGBH-MLA/ams/pull/755#issuecomment-1680731291) which probably impacts your work here too.

ShanaLMoore (approved):

PR #762, opened 2023-08-16
Add env variables to rails 6 context

ShanaLMoore:
We noticed that rails 6 tests are not using the correct gem versions.

ref: https://github.com/WGBH-MLA/ams/pull/755#issuecomment-1680731291
ref: [solution to selenium 404](https://stackoverflow.com/questions/70381644/how-to-resolve-failure-error-raise-errorservererror-self-seleniumwebdriver)

We expect rails 6 to report 10 failures and to use the correct gem versions found in Gemfile_next.lock.

With this PR, the rails 6 context is correctly using the update gem versions: 

image

Additionally, I've confirmed that there's a total of 10 failing feature specs across all containers: 

image


ShanaLMoore:
TODO: Mark the feature specs as pending.

orangewolf (approved):

PR #769, opened 2023-08-18
stop doing the yarn check

orangewolf:
its removed in most Rails versions

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -6,7 +6,7 @@ Rails.application.configure do
   config.cache_classes = !!Sidekiq.server?
 
   # Verifies that versions and hashed value of the package contents in the project's package.json
-  config.webpacker.check_yarn_integrity = true
+  config.webpacker.check_yarn_integrity = false
   # Method for using environment variables for Booleans
   def truthy_env_var?(val)
     ['yes', 'true', '1'].include? val.to_s.downcase.strip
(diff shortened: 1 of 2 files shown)


laritakr (approved):

config/environments/development.rb
```diff
@@ -6,7 +6,7 @@
   config.cache_classes = !!Sidekiq.server?
 
   # Verifies that versions and hashed value of the package contents in the project's package.json
-  config.webpacker.check_yarn_integrity = true
+  config.webpacker.check_yarn_integrity = false
```
ShanaLMoore, line 9: Should we add the dual boot conditional to if? 

`config.webpacker.check_yarn_integrity = (App.rails_5_1? ? true : false)`

config/environments/test.rb
```diff
@@ -14,6 +14,8 @@
   # and recreated between test runs. Don't rely on the data there!
   config.cache_classes = true
 
+  config.webpacker.check_yarn_integrity = false
```
ShanaLMoore, line 17: Should we add the dual boot conditional to if? 

`config.webpacker.check_yarn_integrity = (App.rails_5_1? ? true : false)`

config/environments/development.rb
orangewolf: it should always be false. it has to do with a buggy yarn command

PR #781, opened 2023-08-23
:gear: add sentry-ruby to gemfile

ShanaLMoore:
I previously set the SENTRY_DSN for each environment but did not notice sentry wasn't part of the Gemfile. Our other projects use sentry-raven which is deprecated, so this commit installs [sentry-ruby](https://github.com/getsentry/sentry-ruby) to track errors in different enviornments.

Follow up for Issue:
- https://github.com/scientist-softserv/ams/issues/76

ref: 
- https://github.com/scientist-softserv/ams/issues/89

Sentry-ruby docs [recommends](https://github.com/getsentry/sentry-ruby#getting-started) installing the following. I added all but delayed_job and resque since we use sidekiq instead. I also excluded "sentry-opentelemetry". At a minimum, sentry-ruby and sentry-rails is required.

```ruby
gem "sentry-rails"
gem "sentry-sidekiq"
gem "sentry-delayed_job"
gem "sentry-resque"
gem "sentry-opentelemetry"
```

Do we need a config file? (tested locally, it seems like we do)

```ruby
# config/initializer/sentry.rb
Sentry.init do |config|
  config.dsn = ENV['SENTRY_DSN']
  config.environment = ENV['SENTRY_ENVIRONMENT']
end

```

I was able to successfully trigger and capture an error from my local environment after doing so: 

image

# Notes

FYI: performance monitoring and tracing systems like Sentry, the term ["traces sample rate"](https://docs.sentry.io/platforms/javascript/performance/#configure-the-sample-rate) refers to the proportion or percentage of requests or transactions that are actually captured and sent to the monitoring system for analysis. It's a way to manage the amount of data being collected and transmitted while still gaining insights into the performance of your application.

For example, if you have a traces sample rate of 10%, it means that only 10% of the requests your application receives will have their performance data (traces) collected and sent to the monitoring system like Sentry. The remaining 90% of requests will not be captured, reducing the overall data volume and potential impact on your application's performance and network resources.

A common approach could be to set the traces sample rate in staging to around 30-50% initially, and then monitor the performance impact. You can then adjust the sample rate up or down based on the factors (like: performance impact, budget, traffic volume, debugging needs, and resource usage) and how well your staging environment can handle the additional load.

So in staging I'm setting it to 100% for now. We can monitor and adjust the env variable. In prod I set it to 10% as requested by Rob.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -98,6 +98,12 @@ group :development do
   # gem 'spring-watcher-listen', '~> 2.0.0'
   gem "letter_opener"
   gem 'faker'
+  # Sentry-ruby for error handling
+  gem "sentry-ruby"
+  gem "sentry-rails"
+  gem "sentry-sidekiq"
+  gem "sentry-resque"
+  gem "sentry-opentelemetry"
   # gem 'xray-rails' should be commented out when actively using sidekiq.
 end
 
@@ -123,12 +129,6 @@ gem 'database_cleaner'
 gem 'redlock', '~> 1.0'
 gem 'httparty', '~> 0.21'
 
-# Sentry-ruby for error handling
-gem "sentry-ruby"
-gem "sentry-rails"
-gem "sentry-sidekiq"
-gem "sentry-resque"
-gem "sentry-opentelemetry"
 # Adding pry to all environments, because it's very useful for debugging
 # production environments on demo instances.
 gem 'pry-byebug', platforms: [:mri, :mingw, :x64_mingw]
--- a/Gemfile
+++ b/Gemfile
@@ -123,7 +123,7 @@ gem 'database_cleaner'
 gem 'redlock', '~> 1.0'
 gem 'httparty', '~> 0.21'
 
-# Sentry-raven for error handling
+# Sentry-ruby for error handling
 gem "sentry-ruby"
 
 # Adding pry to all environments, because it's very useful for debugging
--- a/Gemfile
+++ b/Gemfile
@@ -127,7 +127,6 @@ gem 'httparty', '~> 0.21'
 gem "sentry-ruby"
 gem "sentry-rails"
 gem "sentry-sidekiq"
-gem "sentry-resque"
 
 # Adding pry to all environments, because it's very useful for debugging
 # production environments on demo instances.
--- a/Gemfile
+++ b/Gemfile
@@ -98,11 +98,6 @@ group :development do
   # gem 'spring-watcher-listen', '~> 2.0.0'
   gem "letter_opener"
   gem 'faker'
-  # Sentry-ruby for error handling
-  gem "sentry-ruby"
-  gem "sentry-rails"
-  gem "sentry-sidekiq"
-  gem "sentry-resque"
   # gem 'xray-rails' should be commented out when actively using sidekiq.
 end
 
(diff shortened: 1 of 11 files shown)


Gemfile
```diff
@@ -123,6 +123,9 @@ gem 'database_cleaner'
 gem 'redlock', '~> 1.0'
 gem 'httparty', '~> 0.21'
 
+# Sentry-raven for error handling
```
ShanaLMoore:
```suggestion
# Sentry-ruby for error handling
```

Gemfile
```diff
@@ -123,6 +123,12 @@ gem 'database_cleaner'
 gem 'redlock', '~> 1.0'
 gem 'httparty', '~> 0.21'
 
+# Sentry-ruby for error handling
+gem "sentry-ruby"
+gem "sentry-rails"
+gem "sentry-sidekiq"
+gem "sentry-resque"
```
orangewolf: why resque? we are only using sidekiq here I beleive

config/initializers/sentry.rb
```diff
@@ -0,0 +1,17 @@
...
+# in Rails, this might be in config/initializers/sentry.rb
+Sentry.init do |config|
+  config.dsn = ENV['SENTRY_DSN']
+  config.environment = ENV['SENTRY_ENVIRONMENT']
+  config.breadcrumbs_logger = [:active_support_logger, :http_logger]
+
+  # Set traces_sample_rate to 1.0 to capture 100%
+  # of transactions for performance monitoring.
+  # We recommend adjusting this value in production.
+  config.traces_sample_rate = 1.0
```
orangewolf: this should be 0.1 for prod please

orangewolf (changes requested):

config/initializers/sentry.rb
ShanaLMoore: hard coded or as an environment variable? I'll make it an env unless otherwise stated, so we can easily change it live if need be.

Gemfile
ShanaLMoore: Thanks for confirming. I got its description mixed up with sentry-ruby and didn't realize it was another [background worker](https://github.com/resque/resque) at the time. I'll remove it. 


Resque is a Redis-backed Ruby library for creating background jobs, placing them on multiple queues, and processing them later.

ShanaLMoore:
FYI: performance monitoring and tracing systems like Sentry, the term "traces sample rate" refers to the proportion or percentage of requests or transactions that are actually captured and sent to the monitoring system for analysis. It's a way to manage the amount of data being collected and transmitted while still gaining insights into the performance of your application.

For example, if you have a traces sample rate of 10%, it means that only 10% of the requests your application receives will have their performance data (traces) collected and sent to the monitoring system like Sentry. The remaining 90% of requests will not be captured, reducing the overall data volume and potential impact on your application's performance and network resources.

A common approach could be to set the traces sample rate in staging to around 30-50% initially, and then monitor the performance impact. You can then adjust the sample rate up or down based on the factors (like: performance impact, budget, traffic volume, debugging needs, and resource usage) and how well your staging environment can handle the additional load.

So in staging I'm setting it to 100% for now. We can monitor and adjust the env variable. In prod I set it to 10% as requested by Rob.

laritakr (approved):

orangewolf (approved):

PR #782, opened 2023-08-23
🐛 Timeout when creating new work via UI

laritakr:
🐛 Override AdminSetCreateService

refs 
- https://github.com/samvera/hyrax/issues/6171
- https://github.com/scientist-softserv/ams/issues/85

Finding the default admin set ID results in a timeout for large admin sets. Override to use the first admin set until Hyrax ticket above is completed.

🎁 Edit fields form exception workaround

Add try to view to prevent no-method exception on hidden? `f.object.class.try(:hidden?, key)`

### DIFF
--- a/app/services/hyrax/admin_set_create_service_decorator.rb
+++ b/app/services/hyrax/admin_set_create_service_decorator.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+# OVERRIDE Hyrax 4.0.0 AdminSetCreateService to return first admin set (default)
+# until timeout issue is resolved
+module Hyrax
+  module AdminSetCreateServiceDecorator
+    private
+
+    def find_default_admin_set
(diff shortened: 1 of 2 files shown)


kirkkwang:
✅ QA'd on demo and passes

image

kirkkwang (dismissed):

app/services/hyrax/admin_set_create_service_decorator.rb
```diff
@@ -0,0 +1,20 @@
...
+# until timeout issue is resolved
+module Hyrax
+  module AdminSetCreateServiceDecorator
+    private
+
+    def find_default_admin_set
+      if App.rails_5_1?
+        super
+      else
+debugger
```
kirkkwang: ah ha!

kirkkwang (approved):

PR #786, opened 2023-08-25
I93 bulkrax upgrade

ShanaLMoore:
REF: https://github.com/scientist-softserv/ams/issues/93

This PR upgrades bulkrax from 1.0.2 to 5.3.0 but leaves GBH's custom relationship code so that the XML importer can form child relations.

# XML PARSER
- [x] successfully imports data and forms/displays relationships (using their existing custom relationship code)

image

image

# XML/MANIFEST PARSER
Marked as low priority so we can make a separate ticket for this

# CSV PARSER
Marked as the lowest priority so we can make a separate ticket for this


app/views_rails_6_1/bulkrax/entries/show.html.old.erb
```diff
@@ -1,4 +1,5 @@
 <%# OVERRIDE BULKRAX 1.0.2 to display ID instead of record class name %>
+<% rails 'hel' %>
```
ShanaLMoore: oops

Gemfile
```diff
@@ -21,8 +21,7 @@ if ENV['DEPENDENCIES_NEXT'] && !ENV['DEPENDENCIES_NEXT'].empty?
   gem 'twitter-typeahead-rails', '0.11.1.pre.corejavascript'
   group :bulkrax do
```
kirkkwang: i'm thinking we don't need this group anymore since it's just bulkrax now

ShanaLMoore: true

ShanaLMoore:
converting back to draft to resolve failing 5.1 spec. the 6.1 specs can be handled in their respective tickets.

kirkkwang (approved):

PR #789, opened 2023-08-28
Valkyrie models

kirkkwang:
# Story

Ref:
- https://github.com/scientist-softserv/ams/issues/36

Generate valkyrie models of the following:
- Asset
- PhysicalInstantiation
- DigitalInstantiation
- EssenceTrack
- Contribution

Now we have:
- AssetResource
- PhysicalInstantiationResource
- DigitalInstantiationResource
- EssenceTrackResource
- ContributionResource

# Notes

Even after registering the test_adapter I was getting still getting errors so I'm skipping all the generated model test for now.
https://github.com/samvera/hyrax/blob/4.x-stable/spec/test_app_templates/lib/generators/test_app_generator.rb#L219-L228

### DIFF
--- a/config/metadata/asset_resource.yaml
+++ b/config/metadata/asset_resource.yaml
@@ -1,0 +1,22 @@
+# Simple yaml config-driven schema which is used to define model attributes,
+# index key names, and form properties.
+#
+# Attributes must have a type but all other configuration options are optional.
+#
+# attributes:
+#   attribute_name:
+#     type: string
+#     multiple: false
+#     index_keys:
+#       - "attribute_name_sim"
+#     form:
+#       required: true
+#       primary: true
+#       multiple: false
+#
+# @see config/metadata/basic_metadata.yaml for an example configuration
+#
+# Generated via
+#  `rails generate hyrax:work_resource AssetResource`
+
+attributes: {}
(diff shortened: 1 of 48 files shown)


ShanaLMoore:
Should we upgrade Hyrax too or has that already been done?

kirkkwang:
That's gonna be tricky with our dual boot situation, Hyrax `main` depends on Ruby 3.2 or higher.

ShanaLMoore:
> That's gonna be tricky with our dual boot situation, Hyrax `main` depends on Ruby 3.2 or higher.

Oh right, we put a pause on that for a moment. Hmm 🤔 This looks fine for me but wonder if we should get Rob's eyes on it.

config/metadata/asset_resource.yaml
```diff
@@ -0,0 +1,357 @@
...
+    type: string
+    multiple: true
+    index_keys:
+      - "series_description_tesim"
+    form:
+      required: false
+      primary: false
+      multiple: true
+  intended_children_count:
+    type: string
```
laritakr, line 343: This needs to include index_keys

kirkkwang: additional indexing will be handled in another issue

orangewolf (approved):

PR #795, opened 2023-08-29
I75 cleanup upgrade

ShanaLMoore:
In this PR we remove all support for rails 5.1

the pipeline is failing but we're hoping we can merge w Rob's branch instead of making duplicate changes to remove the rails 5 build/tests


ShanaLMoore:
LaRita confirmed that we can connect to the browser and run the specs. 

All specs passed locally.

orangewolf (approved):

PR #796, opened 2023-08-30
:broom: restore view directory

ShanaLMoore:
In a previous commit where we removed support for rails 5.1 we forgot to copy over the view directory, causing styling reversion. Now the UI looks better. 

## BEFORE

image (31)

## AFTER

image


kirkkwang (approved):
looks good!

PR #803, opened 2023-09-15
add listener for save to essentracks, instantiations and assets.

orangewolf:
validate aapb status when called.


app/jobs/bulkrax/child_relationships_job.rb
```diff
@@ -42,6 +42,7 @@ def work_membership
       child_entries.each do |child_entry|
         child_record = child_entry.factory.find
         next if child_record.blank? or child_record.is_a?(Collection)
+        next if parent_record.blank?
```
laritakr, line 45: You didn't begin work using the most recent code. I had pushed a different change for the child relationships job to handle parent record being blank which is in the ruby tuesday branch, and makes this unnecessary.

PR #811, opened 2023-12-12
bring back assets to search

orangewolf:
and make sure they do not change has_model_ssim as it reindexes


bkiahstroud (approved):
Assuming the failing specs aren't related to these changes, this looks good to me

PR #813, opened 2023-12-15
Fix indexing of children

orangewolf:
assets should get certain fields from their children added to themselves when indexing


jeremyf (approved):

config/metadata/.#physical_instantiation_resource.yaml
```diff
@@ -0,0 +1 @@
+rob@Robs-MacBook-Pro.local.83041
```
bkiahstroud, line 1: I'm guessing this got committed by accident?

bkiahstroud (changes requested):

PR #817, opened 2023-12-20
Upgrade Hyrax Batch Ingest to Valkyrie

orangewolf:


app/models/hyrax/pcdm_collection_decorator.rb
```diff
@@ -1,5 +1,6 @@
-# OVERRIDE Hyrax 5.0 to add basic metadata to collection
+ # OVERRIDE Hyrax 5.0 to add basic metadata to collection
```
alishaevn:
```suggestion
# OVERRIDE Hyrax 5.0 to add basic metadata to collection
```

alishaevn (dismissed):
idk the context behind this, but from a "does this code work" perspective, it looks fine to me.

alishaevn (dismissed):

alishaevn (dismissed):

PR #824, opened 2024-01-25
make sure xml gets added to digital instantiations

orangewolf:


orangewolf:
the comma here mean the case statement was checking 2 things and doing nothing instead of checking one thing and then doing one thing if it is true

jpengst (approved):

ShanaLMoore (approved):

PR #825, opened 2024-02-02
Simple form fix

orangewolf:


lib/reprocessor.rb
```diff
@@ -152,6 +152,19 @@ def with_id_lines
...
+  def lambda_create_relationships
+    @lambda_create_relationships ||= lambda { |line, progress|
+      id = line.strip
+      e = Bulkrax::Entry.find(id)
+      ::SEEN ||= []
+      if ::SEEN.include?(e.importer.id)
+        e.parser.create_parent_child_relationships
+      else
+        ::SEEN << e.importer.id
+      end
```
bkiahstroud: Can you help me understand this conditional? It looks like we're skipping creating relationships for the first entry from the importer

orangewolf: That’s because it’s backwards. It should be !include?  Good catch

bkiahstroud (approved):

PR #834, opened 2024-02-22
i121-script-to-move-records

aprilrieger:
First pass at adding the script, need to make sure the paths work on demo aka prod env

in your terminal enter a screen process with `screen -xRR`

You will need to transfer the txt file from the client to the worker pod in the k8 cluster
`k cp /Users/aprilrieger/softserv/ams/NJN_forupdates_txt.txt worker-pod-name:/tmp/imports/NJN_forupdates_txt.txt`

Then drop into a rails console `rails c`

Then run the job and pass the path of the file: `PriorityImportsJob.perform_now('tmp/imports/NJN_forupdates_txt.txt')`

Once the job is completed you will need to create an importer from a path and add the path to the folder (the name of the filename, example: `NJN_forupdates_txt`)

### DIFF
--- a/app/jobs/process_files_job.rb
+++ b/app/jobs/process_files_job.rb
@@ -1,0 +1,42 @@
+# app/jobs/process_files_job.rb
+class ProcessFilesJob < ApplicationJob
+  queue_as :default
+
+  def perform(file_path)
+    puts "file_path=#{file_path}"
+    # Extract the directory where the file is located and the base name without .txt extension
+    dir_path = File.dirname(file_path)
+    puts "dir_path=#{dir_path}"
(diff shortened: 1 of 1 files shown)


app/jobs/process_files_job.rb
```diff
@@ -0,0 +1,42 @@
+# app/jobs/process_files_job.rb
+class ProcessFilesJob < ApplicationJob
+  queue_as :default
+
+  def perform(file_path)
+    puts "file_path=#{file_path}"
```
aprilrieger, line 6: Note to self: remove all the comments once fine-tuned.

PR #839, opened 2024-03-06
Update annotation_types.yml

afred:
Adds "Note" as a valid annotation type.


ekemeyer:
duplicate - going with passing PR #971

Issue #831, opened 2024-02-13
Fix Sony CI linking button in Edit interfae


afred:
### Because

The front-end button an Edit Asset pages that links Sony Ci records to AMS Assets is broken

### Done when

As an admin user
When I am looking at an AMS Asset that has a corresponding Sony Ci record (with the AAPB ID in the name) but it not currently linked to the Sony Ci record with metadata
And when I click the link to find Sony Ci records
It sucessfully finds all the corresponding Sony Ci records and populates the text boxes with the correct Sony Ci IDs
So that I can save it and link the AMS2 Asset with the Sony Ci media.

### Additional context

I think this is due to a Javascript error, which is erroring because the javascript code is relying on old URLs that are no longer valid given the new URL structure that comes from the new models names, i.e. use `/concerns/asset_resource` instead of `/concern/asset`

PR #840, opened 2024-03-06
Fixes JS selector to find Sony Ci media from Asset Edit interface

bkiahstroud:
Rename selector to properly find elements. 

Related:
- #831 
- #838


afred (approved):
👍

PR #844, opened 2024-03-07
i123 Index Bulkrax identifier to speed up query

bkiahstroud:
Ref:
- https://github.com/scientist-softserv/ams/issues/123

Before this change, Hyrax::CustomQueries::FindByBulkraxIdentifier would take 90+ seconds to run on production which currently has 1.75 million works.


bkiahstroud:
Duplicate of https://github.com/WGBH-MLA/ams/pull/841

PR #845, opened 2024-03-07
i123 Updated db schema

bkiahstroud:
Ref: 
- https://github.com/scientist-softserv/ams/issues/123
- #841 

Reflects migration added by #841

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2023_08_30_155065) do
+ActiveRecord::Schema.define(version: 2024_03_07_194231) do
 
   # These are extensions that must be enabled in order to support this database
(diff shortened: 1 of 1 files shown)


afred:
@bkiahstroud sorry, I'm not really familiar with the syntax there. Looks like you're adding an index to bulkrax identifier, which makes sense, but what is the logic/meaning behind the `->` and `->>` and the `where` condition?

bkiahstroud:
> @bkiahstroud sorry, I'm not really familiar with the syntax there. Looks like you're adding an index to bulkrax identifier, which makes sense, but what is the logic/meaning behind the `->` and `->>` and the `where` condition?

The metadata column is a JSON field, and this index is specifically targeting the `bulkrax_identifier` key within that column. The `->` and `->>` syntax is SQL query syntax for JSON fields. As for the `WHERE`, that is saying that the index only includes rows where `bulkrax_identifier` is not null

afred (approved):
👍

PR #848, opened 2024-03-11
Adds script to auto start importer

K8Sewell:
# Summary
Adds a rake task to check for running import jobs and if none are running will trigger another import job to be run.

# Related Ticket
[i124](https://github.com/scientist-softserv/ams/issues/124)

# Co-Authored By:
* JP - @jpengst
* Amanda - @akfosterdev

### DIFF
--- a/lib/tasks/ams/trigger_importer.rake
+++ b/lib/tasks/ams/trigger_importer.rake
@@ -1,0 +1,41 @@
+require 'ams'
+
+
+# only query for default and import queues - ignore all others
+namespace :ams do
+  desc 'Resets Fedora, Solr, Database, and creates default admin set'
+  task trigger_importer: :environment do
+    # queue if an importer is running
+
+    # if not then queue the next importer
+
+
+    # If using this to determine the next importer to run, you'll need to adjust this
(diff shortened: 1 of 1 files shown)


lib/tasks/ams/trigger_importer.rake
```diff
@@ -0,0 +1,26 @@
+require 'ams'
```
bkiahstroud: Why is this line here?

bkiahstroud (commented):
I'd love to see some logging, even if it's super bare bones. Would y'all mind if I just implement this myself? I can do it pretty quick

lib/tasks/ams/trigger_importer.rake
K8Sewell: It was in the other rake tasks so we assumed it was necessary but can pull if it's not.

K8Sewell:
> I'd love to see some logging, even if it's super bare bones. Would y'all mind if I just implement this myself? I can do it pretty quick

Yes, please feel welcome to add.  Sorry that logging slipped our minds when we were mobbing.  Just lmk if it's easier or faster for me to add logging for you.

bkiahstroud (approved):

PR #861, opened 2024-04-12
i16 Expand AssetDestroyer to also destroy AssetResources

bkiahstroud:
## expand AssetDestroyer to also delete AssetResources

ef949bb1b5e5f151882036e2fe6a14bfdee76b5f

## don't set validation status on deleted Asset

ae13cd12b1fe53595fed7eb545013187f848df07

Prevents the following error when deleting an AssetResource via the
Valkyrie Transactions:

`SolrDocument#get_members: undefined method 'members' for AssetResource`

Setting the validation status required looking up the Asset's members
recursively. But since we're in the middle of deleting things:

1. Expected data was missing, and more importantly,
2. We don't care about setting the Asset's validation status because
   it's being deleted

## use transaction to thoroughly delete members

1a0a4a65bce727a873d9cb784aea9bf9b33e1615

The Transaction ensures the members will be deleted thoroughly. For
example, using the transaction means all the related Sipity records will
be destroyed, which is what we want

### DIFF
--- a/app/services/ams/asset_destroyer.rb
+++ b/app/services/ams/asset_destroyer.rb
@@ -1,21 +1,22 @@
 module AMS
   class AssetDestroyer
-    attr_accessor :asset_ids, :user_email
+    attr_accessor :asset_ids, :user_email, :logger
 
     def initialize(asset_ids: [], user_email: nil)
       @asset_ids = Array(asset_ids)
       @user_email = user_email
+      @logger = Logger.new(Rails.root.join('tmp', 'imports', 'asset_destroyer.log'))
     end
 
     def destroy(asset_ids)
-      puts "Initiating destruction sequence for #{asset_ids.count} Assets..."
+      logger.info "Initiating destruction sequence for #{asset_ids.count} Assets..."
       Array(asset_ids).each do |asset_id|
         destroy_asset_by_id asset_id
       end
     end
 
     def eradicate_asset_tombstones(asset_ids)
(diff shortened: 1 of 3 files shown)


app/services/ams/asset_destroyer.rb
```diff
@@ -64,6 +64,8 @@ def destroy_in_postgres(asset_id)
           .with_step_args('work_resource.delete' => { user: user },
                           'work_resource.delete_all_file_sets' => { user: user })
           .call(asset_resource).value!
+      rescue Valkyrie::Persistence::ObjectNotFoundError
+        puts "No AssetResource found with ID #{asset_id}"
```
kirkkwang, line 69: Would a Rails logger be better here than a puts?

bkiahstroud: In principle, yes. I chose to use `puts` in this PR because it follows the established pattern elsewhere in the file. `puts` is also visible when running this in the rails console, whereas I don't think `Rails.logger` is. 

However, a dedicated logger would be the ideal, and I may come back and rework this file to use one. But for the scope of this PR, I think leaving it as `puts` for now is ideal.

bkiahstroud: Since this PR has been here for awhile, I went ahead and implemented the dedicated logger: b620521

afred (dismissed):
👍

afred:
👍

afred (approved):
👍

PR #862, opened 2024-04-16
i16 Process subsets

bkiahstroud:
## various minor improvements

488756504a6b4edb944307e13f30e2b4a579a8ed

## log if file already exists in target dir

5e04b5465be5d4d74b835a018c758f0a6a50b5bc

## add method to audit duplicate XML files

e43449efa3e9a0e08adf5d353ae3afbd5546ef0c

Specifically, we're auditing whether their content differs from each
other or not. If not, we aren't concerned with handling the duplicates
further, but if the content does differ, we need to figure out how

## if XML file doesn't exist, capture and log error

2d09bca1e97336eb2a5a3a20a7ea9bd4e373a6ad

## add methods to delete/recreate subsets

b5ff1033878109d9bc78d0747a832431b0a1208c

## run migrations on test db schema

269a5930cbcf7db24deab630ccef192e22cb33fe

### DIFF
--- a/app/services/ams/asset_destroyer.rb
+++ b/app/services/ams/asset_destroyer.rb
@@ -1,3 +1,5 @@
+require 'ruby-progressbar'
+
 module AMS
   class AssetDestroyer
     attr_accessor :asset_ids, :user_email, :logger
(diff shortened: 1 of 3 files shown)


aprilrieger (dismissed):

afred (approved):
👍

PR #889, opened 2024-06-06
Refactors transaction step for saving admin data

### LINKED ISSUE #882: Extra Sony Ci ID field created on manual update
**Describe the bug**
When a record is updated manually, a new Sony Ci ID attribute is created (even if one already exists with a value in it). This creates a second, broken streaming player in AAPB.

**To Reproduce**
1. Go to any record with an existing Sony Ci value
2. Manually update the record
3. See new Sony Ci field appear
4. Save record and push to AAPB. See second streaming video

**Expected behavior**
Maybe give user the option to Add a sony ci ID, but don't force it to be generated/triggered by the act of manually updating a record

**Screenshots**
Screenshot 2024-05-21 at 3 14 08 PM

**Additional context**
Staff have historically made a mental note to remove this before completing the update, however not all staff are/were aware it was happening or staff forget because it's an unusual thing to have to remember.

afred:
By default, multi-valued input fields display an additional set of input fields that are blank. These get submitted by the form and are usually discarded, but because we are treating Sony Ci ID field special, it was erroneously saving additional blank values for Sony Ci ID, which was resulting in multiple players of the same media dislaying on AAPB when the record was published.

Multiple Sony Ci IDs are allowed in cases of mult-part Assets. They are stored as a serialized JSON array in the AdminData.sonyci_id field, which is related to Asset reords via a global ID field on AdminData.gid. And the global ID fields was used instead of a normal foreign key because the AssetResource used to be just Asset, and stored in Fedora, not the database, and a global ID was the closest thing to a foreign key across 2 persistence layers that we could think of :)

Fixes #882.

### DIFF
--- a/app/services/aapb/batch_ingest/csv_item_ingester.rb
+++ b/app/services/aapb/batch_ingest/csv_item_ingester.rb
@@ -213,8 +213,8 @@ module AAPB
         def set_admin_data_attributes(admin_data, attributes)
           # add existing admin_data values so they're preserved in the AssetActor
           AdminData.attributes_for_update.each do |admin_attr|
-            next unless attributes.keys.include?(admin_attr.to_s)
-            admin_data.send("#{admin_attr}=", attributes[admin_attr.to_s])
+            next unless attributes.keys.include?(admin_attr)
+            admin_data.send("#{admin_attr}=", attributes[admin_attr])
           end
         end
 
--- a/app/transactions/ams/steps/create_aapb_admin_data.rb
+++ b/app/transactions/ams/steps/create_aapb_admin_data.rb
@@ -58,6 +58,8 @@ module Ams
             change_set.fields[field].reject {|v| v.blank? }
           elsif change_set.fields[field].blank?
             nil
+          else # non-multiple with value present
+            change_set.fields[field]
           end
 
           if can_empty_field?(field) || new_admin_data_value.present?
--- a/app/transactions/ams/steps/create_aapb_admin_data.rb
+++ b/app/transactions/ams/steps/create_aapb_admin_data.rb
@@ -64,6 +64,8 @@ module Ams
 
           if can_empty_field?(field) || new_admin_data_value.present?
             admin_data.write_attribute(field, new_admin_data_value)
+          else
+            Rails.logger.debug("Chose not to update AdminData #{admin_data.id}'s field #{field} with value(s): #{change_set.fields[field]}")
           end
         end
       end
--- a/app/transactions/ams/steps/create_aapb_admin_data.rb
+++ b/app/transactions/ams/steps/create_aapb_admin_data.rb
@@ -56,7 +56,7 @@ module Ams
         if admin_data_values.key?('sonyci_id')
           admin_data_values['sonyci_id'] = Array(admin_data_values['sonyci_id']).reject(&:blank?)
         end
-        admin_data.update!(admin_data_values)
+        admin_data.assign_attributes(admin_data_values)
       end
 
       def delete_removed_annotations(admin_data, change_set)
--- a/app/transactions/ams/steps/create_aapb_admin_data.rb
+++ b/app/transactions/ams/steps/create_aapb_admin_data.rb
@@ -51,35 +51,12 @@ module Ams
       end
 
       def set_admin_data_attributes(admin_data, change_set)
-        AdminData.attributes_for_update.each do |field|
-          field = field.to_s
-          
-          # Conditionally set new value for admin data field
-          new_admin_data_value = if AdminData::SERIALIZED_FIELDS.include?(field.to_sym)
-            # Filter out blanks if the field is a serialized field, e.g. sonyci_id.
-            change_set.fields[field].reject {|v| v.blank? }
-          elsif change_set.fields[field].blank?
-            # If not serialized and blank, convert to nil
-            nil
-          else
-            # Otherwise keep value as is.
-            change_set.fields[field]
-          end
-
-          # If the AdminData field can be emptied OR if we have a non-blank value, then update the AdminData field.
-          if can_empty_field?(field) || new_admin_data_value.present?
-            admin_data.write_attribute(field, new_admin_data_value)
-          else
-            Rails.logger.debug("Chose not to update AdminData #{admin_data.id}'s field #{field} with value(s): #{change_set.fields[field]}")
-          end
+        admin_data_values = change_set.fields.slice(*AdminData.attributes_for_update).compact
+        # If Sony Ci IDs are present, ensure there are no blank ones, such as empty strings.
+        if admin_data_values.key?('sonyci_id')
+          admin_data_values['sonyci_id'] = Array(admin_data_values['sonyci_id']).reject(&:blank?)
         end
-      end
-
-      def can_empty_field?(field)
-        %w(
-          bulkrax_importer_id
-          hyrax_batch_ingest_batch_id
-        ).exclude?(field.to_s)
+        admin_data.update!(admin_data_values)
       end
 
       def delete_removed_annotations(admin_data, change_set)
--- a/app/transactions/ams/steps/create_aapb_admin_data.rb
+++ b/app/transactions/ams/steps/create_aapb_admin_data.rb
@@ -51,22 +51,26 @@ module Ams
       end
 
       def set_admin_data_attributes(admin_data, change_set)
-        AdminData.attributes_for_update.each do |k|
-          # Some attributes are serialized on AdminData, so always send an array
-          k_string = k.to_s
-          if should_empty_admin_data_value?(k, change_set)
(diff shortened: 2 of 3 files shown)


afred:
@bkiahstroud and/or @orangewolf could I get an extra set of 👁️ on the logic here? This is some consequential code as it handles not only our Asset relations to Sony Ci media, but the Asset relation to the Bulkrax Importers and Batch Ingest Batches as well.

I think i tested all possibilities of adding/removing Sony Ci IDs, but did not do any kind of regression testing on persistence of bulkrax identifier or other admin data fields.

app/transactions/ams/steps/create_aapb_admin_data.rb
```diff
@@ -51,22 +51,26 @@ def save_aapb_admin_data(change_set)
...
-          # Some attributes are serialized on AdminData, so always send an array
-          k_string = k.to_s
-          if should_empty_admin_data_value?(k, change_set)
-            AdminData::SERIALIZED_FIELDS.include?(k) ? admin_data.send("#{k}=", Array.new) : admin_data.send("#{k}=", nil)
-          elsif change_set.fields[k_string].present?
-            AdminData::SERIALIZED_FIELDS.include?(k) ? admin_data.send("#{k}=", Array(change_set.fields[k_string])) : admin_data.send("#{k}=", change_set.fields[k_string].to_s)
+        AdminData.attributes_for_update.each do |field|
+          field = field.to_s
+          # Convert emtpy strings to nil, whether the value is a scalar or an array
+          new_admin_data_value = if change_set.fields[field].respond_to?(:select)
```
bkiahstroud: Keying off of `respond_to?(:select)` will likely cast a much wider net than `SERIALIZED_FIELDS` does. Do we understand which fields will be impacted and how they will respond to this change?

app/transactions/ams/steps/create_aapb_admin_data.rb
```diff
@@ -51,22 +51,26 @@ def save_aapb_admin_data(change_set)
...
+            change_set.fields[field].reject {|v| v.blank? }
+          elsif change_set.fields[field].blank?
+            nil
+          end
```
bkiahstroud: Can you confirm if I'm reading this conditional correctly? 

```
if can be multiple
  filter out blank values
else if single and blank
  set new value to nil
end
```

If I am reading that correctly, I think we're missing a scenario, which would be "is single and **not** blank". In that case, I think we just lose the value. I think something like this would resolve this: 

```suggestion
          new_admin_data_value = if change_set.fields[field].respond_to?(:select)
            change_set.fields[field].reject {|v| v.blank? }
          elsif change_set.fields[field].blank?
            nil
          else # non-multiple with value present
            change_set.fields[field]
          end
```

app/transactions/ams/steps/create_aapb_admin_data.rb
```diff
@@ -51,22 +51,26 @@ def save_aapb_admin_data(change_set)
...
+
+          if can_empty_field?(field) || new_admin_data_value.present?
+            admin_data.write_attribute(field, new_admin_data_value)
```
bkiahstroud: Since we only write conditionally, I'd love to see an `else` which logs fields we choose _not_ to write. Historically, this has helped immensely when trying to track down issues along the lines of "I see my data going in but not coming out" 

```suggestion
          if can_empty_field?(field) || new_admin_data_value.present?
            admin_data.write_attribute(field, new_admin_data_value)
          else
            Rails.logger.debug("Chose not to update AdminData #{admin_data.id}'s field #{field} with value(s): #{change_set.fields[field]}")
```

bkiahstroud (changes requested):
I left a few questions and suggestions. Also, it looks like some relevant specs are failing in CI

afred:
@bkiahstroud even more of a simplification...

* Explicitly checks for `sonyci_id` field rather than using `AdminData::SERIALIZED_FIELDS`. This is an unnecessary abstraction that was put in place with the anticipation that we'd end up having more than one serialized field. It's been many years, and we have not, so I don't mind getting rid of that little bit of abstraction since that's the only place it's used. Furthermore, I think we'd be inclined to use regular JSON fields rather than serialized fields should an additional need arise.
* Moves the logic for "you cannot delete values for some fields after they've been set" into validation method `AdminData#validate_undeletable_fields`, which is a more appropriate location since the logic should apply no matter where the model is being used for persistence. This eliminates the need for the `should_empty_admin_data_value` method and reduces logic complexity in `set_admin_data_attributes`
* Removes unnecessary conversion of `AdminData#attributes_for_update` values from string, to symbol, back to string -- just keep them as strings, and we should be good.
* Uses `AdminData#attributes_for_update` to slice out AdminData params from the `AssetResourceForm` rather than looping over it, allowing us to update the

afred:
Specs are passing for me locally, so idk what's up with the failures here. Can you replicate the failures locally?

bkiahstroud:
> Specs are passing for me locally, so idk what's up with the failures here. Can you replicate the failures locally?

I can, I get the same failures locally

app/services/aapb/batch_ingest/csv_item_ingester.rb
```diff
@@ -213,8 +213,8 @@ def set_batch_ingest_id_on_related_asset(work_id, ability)
         def set_admin_data_attributes(admin_data, attributes)
           # add existing admin_data values so they're preserved in the AssetActor
           AdminData.attributes_for_update.each do |admin_attr|
-            next unless attributes.keys.include?(admin_attr.to_s)
-            admin_data.send("#{admin_attr}=", attributes[admin_attr.to_s])
```
bkiahstroud, line 217: Why remove the `#to_s`'s here? If I put them back, one of the three failing specs pass

bkiahstroud:
@afred do you have changes locally that haven't been pushed up to this PR yet? For example, I don't see a new method named `AdminData#validate_undeletable_fields` that you referenced in [this comment](https://github.com/WGBH-MLA/ams/pull/889#issuecomment-2165869946)

app/services/aapb/batch_ingest/csv_item_ingester.rb
afred: it shouldn't be necessary. I found that they start as strings, get converted to symbols, but then in all comparisons (except maybe in a spec somewhere) they are compared as strings. I'm guessing the symbols were used thinking it would be a best practice to store field names as an array of symbols rather than strings.. but if strings are actually what's being used everywhere, it's not a best practice anymore.

app/transactions/ams/steps/create_aapb_admin_data.rb
```diff
@@ -51,22 +51,12 @@ def save_aapb_admin_data(change_set)
...
       def set_admin_data_attributes(admin_data, change_set)
-        AdminData.attributes_for_update.each do |k|
-          # Some attributes are serialized on AdminData, so always send an array
-          k_string = k.to_s
-          if should_empty_admin_data_value?(k, change_set)
-            AdminData::SERIALIZED_FIELDS.include?(k) ? admin_data.send("#{k}=", Array.new) : admin_data.send("#{k}=", nil)
-          elsif change_set.fields[k_string].present?
-            AdminData::SERIALIZED_FIELDS.include?(k) ? admin_data.send("#{k}=", Array(change_set.fields[k_string])) : admin_data.send("#{k}=", change_set.fields[k_string].to_s)
-          end
+        admin_data_values = change_set.fields.slice(*AdminData.attributes_for_update).compact
```
bkiahstroud, line 54: `#slice` only works if the keys are the same type. We know `AdminData.attributes_for_update` are strings, is `change_set.fields`?

app/transactions/ams/steps/create_aapb_admin_data.rb
```diff
@@ -51,22 +51,12 @@ def save_aapb_admin_data(change_set)
...
+          admin_data_values['sonyci_id'] = Array(admin_data_values['sonyci_id']).reject(&:blank?)
         end
-      end
-
-      def should_empty_admin_data_value?(key, change_set)
-        return false if %i[bulkrax_importer_id hyrax_batch_ingest_batch_id].include?(key)
-
-        # The presence of the key with a "blank" value indicates we're intentionally emptying the value
-        change_set.fields.key?(key) && change_set.fields[key].blank?
+        admin_data.update!(admin_data_values)
```
bkiahstroud: `#update!` will persist the changes to the database, but I think the purpose of this method is just to set the values. The changes will be persisted by `#save_aapb_admin_data`. I recommend using `#assign_attributes`

app/transactions/ams/steps/create_aapb_admin_data.rb
afred: Yeah, pretty sure from debugging the specs. Also the previous comparison assumed string keys: `change_set.fields[k_string].present?`

bkiahstroud (approved):

PR #894, opened 2024-06-28
only add bulkrax files if bulkrax is enabled

bkiahstroud:


bkiahstroud:
A similar change to this was implemented in another PR: 

- #911

PR #907, opened 2024-08-22
Update annotation_types.yml

ekemeyer:
Need to add a new annotation to account for annotations in New Jersey Network that do not currently have a defined type. Spoke with Rebecca and confirmed new annotation term/id.


afred (dismissed):
👍

bkiahstroud (commented):
The reported spec failures are legitimate. The failing specs point to [this line](https://github.com/WGBH-MLA/ams/blame/develop/app/indexers/asset_resource_indexer.rb#L31). 

If you replace `.empty?` with `.blank?` on that line, that should resolve the failing specs.

bkiahstroud (approved):

PR #925, opened 2024-10-01
Adds Fedora record removal to DeletePhysicalInstantiations fix

afred:


app/models/concerns/ams/work_behavior.rb
```diff
@@ -18,14 +18,24 @@ def to_rdf_representation
...
-        rescue Valkyrie::Persistence::ObjectNotFoundError
-          Rails.logger.warn("Could not find member #{id} for #{self.id}")
-        end
+      @members ||= []
+      member_id_vals = member_ids.map { |id| id.to_s }.to_set
+      ids_from_members = @members.map { |m| m.id.to_s }.to_set
+      # If the member_ids do not match the IDs within the members, then re-set the members to be an accurate reflection of the member IDs.
+      # This allows us to modify resource.member_ids and have #members accurately reflect the changes.
+      if member_id_vals != ids_from_members
+        @members = member_ids.map do |id|
```
bkiahstroud, line 27: Can you explain more about what's happening on these lines and why?

lib/fix/delete_asset_resources.rb
```diff
@@ -3,20 +3,68 @@
...
+    def results; @results ||= []; end
+
+    def work_destroy_transaction
+      Hyrax::Transactions::WorkDestroy.new
+    end
+
+    def destroy_work_and_members(resource:)
+      result = { resource: resource }
+      resource.members.each do |member|
+        resource.members.delete(member)
```
bkiahstroud: Can you please explain the purpose of this line? (28)

lib/fix/sample_data.rb
```diff
@@ -0,0 +1,87 @@
...
+    @rand_start ||= rand(0..(total_assets - asset_sample_size - 1))
+  end
+
+  def total_assets
+    solr.get('select', params: {q: 'has_model_ssim:Asset', rows: 0})['response']['numFound']
+  end
+
+  def ars(refresh: false)
+    # Rest instance variables if refresh is true
+    @ars = @ars_with_pids = @ars_with_pis_in_fedora = nil if refresh
```
foglabs: typo on the `@ars_with_pis` variable here, so wouldn't get refreshed

app/models/concerns/ams/work_behavior.rb
afred: The `WorkBehavior#members` is supposed to be a memoized, but accurate, mapping of `member_ids` to objects returned from `Hyrax.query_service`. When deleting members, I remove their IDs from the `member_ids` array of the parent (we could not use the Hyrax transaction for reasons I can explain if you want). 

But then the `#members` would be out of sync. So I fixed the cache-busting condition to just detect if the `@members` and `member_ids` are different, and if so, reset `@members` to accurately reflect `member_ids`.

lib/fix/delete_asset_resources.rb
afred: It's supposed to be removing the child from the parent, but if you are thinking it should be deleting from `member_ids` instead of `members`, I think you are correct.

This `delete_asset_resources` is code I added as a consolidation/replacement for other Asset deleting scripts we have that can be run from command line. This one does a recursive delete to get at EssenceTracks which were being left behind, and I was using it a lot in the course of testing so that's why it's included here. Could be a separate PR but hoping to save a little time by combining. It is stand alone and shouldn't affect the script for deleting physical instantiations.

afred: actually, since this is doing a depth-first recursive call to traverse the tree, removing that line really has no effect -- the everything gets deleted, children first. Removing.

foglabs (approved):

PR #935, opened 2024-11-06
884 batch ingest instantiation reset

afred:


app/services/aapb/batch_ingest/pbcore_xml_ingester_behavior.rb
```diff
@@ -0,0 +1,114 @@
...
+# This module is to DRY up common methods  between PBCoreXMLItemIngester and
+# PBCoreXMLInstantiationReset, but also for any future custom ingester that
+# needs to behave mostly like PBCoreXMLItemIngester.
+module AAPB
+  module BatchIngest
+    module PBCoreXMLIngesterBehavior
+      # Common private methods ingester classes.
+      private
+
+        def batch_item_is_asset?
```
foglabs, line 11: prob not a real problem, but technically this could get confused if one of the pbcore's text nodes contain the string 'pbcoreDescriptionDocument'

docker-compose.yml
```diff
@@ -239,6 +239,8 @@ services:
         condition: service_completed_successfully
     expose:
       - 3000
+    ports:
+      - 3000:3000
```
bkiahstroud: Do you use the `dory` gem during local development? Or do you go to localhost:3000?

afred: `localhost:3000`, but I didn't mean to commit this. Will take it out, thanks!

lib/fix/delete_asset_resources.rb
```diff
@@ -3,20 +3,69 @@
...
+
+    def destroy_work_and_members(resource:)
+      result = { resource: resource }
+
+      resource.members.each do |member|
+        destroy_work_and_members(resource: member)
       end
-      puts "Done."
+
+      log.info "DELETING #{resource.class} #{resource.id.id}..."
```
foglabs, line 32: just flagging to confirm that `resource.id.id` is correct - I love when an id has an id!

foglabs (commented):
looks good! just had two (probably silly) comments

lib/fix/delete_asset_resources.rb
afred: yes, it's kind of silly, but `resource.id` is a Valkyrie::ID or something like that, but instead of having a property called `value` or `val`, it's called `id` again. To add to the confusion in ways only Ruby can, the `Valkyrie::ID` object has a `to_s` method so you _can_ refer to `"#{resource.id}"` when it's being interpolated into a string. 🙃

app/services/aapb/batch_ingest/pbcore_xml_ingester_behavior.rb
afred: yeah, a little risky, but too low to block on this i think. It's not new code, it was just pulled out into a behavior module to it could be included in 2 places. So it's been working without error. But we can make a new ticket for it if needed.

foglabs: nah that's fine, we don't have any records that are themselves about pbcore

foglabs (dismissed):
🎸 !

foglabs (dismissed):
🎸 !

docker-compose.yml
bkiahstroud: Accidentally clicked "Unresolve" 🤦

PR #938, opened 2024-11-19
PI format not required

### LINKED ISSUE #940: remove Format as required for PI
Catalogers would like the format field to not be required anymore for Physical Instantiations to allow for born-digital works that do not have an actual PI.

ekemeyer:
No longer require format for physical instantiations when:

- [x] Using update ingester
- [x] PBCore zipped ingester
- [x] Manually adding or updating a record
- [x] Pushing to AAPB

closes #940


ekemeyer:
Work accomplished with Softserv contract

PR #941, opened 2024-12-09
Update batch_ingest.yml

ekemeyer:
adds title and descriptions without types to list of attributes for multi value add and update ingesters


ekemeyer:
duplicate of #667

Issue #945, opened 2025-01-03
PBCore Zipped nil:NilClass error still ingests asset


ekemeyer:
**Describe the bug**
When using PBCore Zipped ingest, xml from GBH throws the error 
``undefined method `first' for nil:NilClass``

This error happens when children are missing a mediaType value. This is not required for PBCore and so it should accept the xml.

UPDATE: Discussed with Rebecca, Peter, etc. and it was decided that mediaType should remain required (especially if we're moving ahead with not requiring instantiationPhysical (format), so we still need to deal with the secondary issue...

A secondary issue is that instead of ignoring the xml ingest entirely, the asset is ingested and the children are ignored (physical instantiations, digital instantiations).

**To Reproduce**
[See batch 3914 summary](https://ams2.wgbh-mla.org/batches/3914/summary). See any of the 101 assets that failed due to ``undefined method `first' for nil:NilClass`` error. They were all partially ingested into AMS2.

**Expected behavior**
Appropriate and clear error is thrown and nothing gets ingested.

PR #947, opened 2025-01-03
Media type not required

ekemeyer:


ekemeyer:
fixes #945

ekemeyer:
in talking with rebecca we determined we want to keep MediaType as a requirement if we are planning to remove PI Format.

PR #952, opened 2025-01-21
Adds fix script for resaving batches of assets.

### LINKED ISSUE #951: Allow bulk re-save of Assets
### Because
The "missing children" validation status is stuck on lots of records. Re-saving them resets the validation status to OK, as it should be, but currently we can only do that one at a time, and there are many to do. A script for taking in a batch of AAPB IDs and simply re-saving them is pretty trivial, and will save a lot of time.

### Done when
As an AMS2 Admin
When I supply a large list of AAPB IDs to the re-save tool
(which may be a command line tool)
All records are re-saved
In order to update the "missing children" validation status to a value that accurately reflects the state of the record.

afred:
Closes #951

### DIFF
--- a/lib/fix/resave_asset_resources.rb
+++ b/lib/fix/resave_asset_resources.rb
@@ -1,0 +1,53 @@
+require_relative '../../config/environment'
+require_relative 'batch_process'
+
+module Fix
+  class ResaveAssetResources < BatchProcess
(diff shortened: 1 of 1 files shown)


ekemeyer (approved):
hooray, let's do this

PR #971, opened 2025-03-03
Add note as annotationType

ekemeyer:
PR to add "Note" as a valid annotationType _specifically_ for PBCore Zipped ingest. This is to allow 68K minnesota records to be ingested that had multiple invalid annotationTypes that we have merged into a single "Note" annotationType

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -323,6 +323,10 @@ class Asset < ActiveFedora::Base
  def ams1_legacy_metadata
     ams1_legacy_metadata ||= find_annotation_attribute("ams1_legacy_metadata")
   end
+   
+ def note
+    note ||= find_annotation_attribute("note")
+  end
   
   def find_annotation_attribute(attribute)
     if admin_data.annotations.select { |a| a.annotation_type == attribute }.present?
--- a/app/models/asset_resource.rb
+++ b/app/models/asset_resource.rb
@@ -133,6 +133,10 @@ class AssetResource < Hyrax::Work
   def ams1_legacy_metadata
     ams1_legacy_metadata ||= find_annotation_attribute("ams1_legacy_metadata")
   end 
+
+   def note
+    note ||= find_annotation_attribute("note")
+  end
   
   def find_annotation_attribute(attribute)
     if admin_data.annotations.select { |a| a.annotation_type == attribute }.present?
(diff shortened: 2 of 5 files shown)


app/models/asset.rb
```diff
@@ -323,6 +323,10 @@ def proxy_start_time
  def ams1_legacy_metadata
     ams1_legacy_metadata ||= find_annotation_attribute("ams1_legacy_metadata")
   end
+   
+ def note
+    note ||= find_annotation_attribute("note")
+  end
```
afred, line 329: Unindent `end` by one space to match it's opening `def` (same for `ams1_legacy_metadata` above, and any other place you might see where they are unaligned).

app/models/asset_resource.rb
```diff
@@ -133,6 +133,10 @@ def proxy_start_time
   def ams1_legacy_metadata
     ams1_legacy_metadata ||= find_annotation_attribute("ams1_legacy_metadata")
   end 
+
+   def note
+    note ||= find_annotation_attribute("note")
+  end
```
afred, line 139: match indentation of `def` and `end`

afred (commented):
In addition to the indentation fixes... are we sure we want to add this type?

To add an annotation of type "note" seems a little redundant (annotations _are_ notes), and also like it may be leaning into a well known anti-pattern of putting specific things into a generic notes field. This has been an issue with every database I've ever worked with -- when the schema falls short, the notes fields get overused.

There is also the annotations without any value for `annotationType`, which I think was what we were using for adding a generic "note" via `pbcoreAnnotation`. So if we now have a `pbcoreAnnotation` with `annotationType="note"`, we may actually be introducing a 2nd place for putting generic notes.

ekemeyer:
> There is also the annotations without any value for `annotationType`, which I think was what we were using for adding a generic "note" via `pbcoreAnnotation`. So if we now have a `pbcoreAnnotation` with `annotationType="note"`, we may actually be introducing a 2nd place for putting generic notes.

As of right now, at least with PBCore zipped, an annotation without a defined annotationType is not allowed. So we'll need either this or allow for none. Based on the data set I'm working on with Minnesota, either is fine, so I'll close this one out and make a ticket for allowing an annotation without a type.

PR #973, opened 2025-03-04
support annotations without types

ekemeyer:
We have a bunch of records to migrate that do not have an asset annotation type, which is valid PBCore, but not permitted with PBCore Zipped ingester. The alternative is to allow for "Note" as a type and assign all of them that type, however Annotation itself is considered a "Note"

Solves #1026

### DIFF
--- a/app/services/aapb/batch_ingest/pbcore_xml_mapper.rb
+++ b/app/services/aapb/batch_ingest/pbcore_xml_mapper.rb
@@ -48,11 +48,7 @@ module AAPB
         if ENV['SETTINGS__BULKRAX__ENABLED'] == 'true'
           type_id
         else
-          if type_id.present?
-            type_id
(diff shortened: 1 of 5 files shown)


app/services/aapb/batch_ingest/pbcore_xml_mapper.rb
```diff
@@ -26,7 +26,7 @@ def prepare_annotations(annotations)
         final_annotations = []
 
         annotations.each do |anno|
-          annotation_type = find_annotation_type_id(anno.type)
+          annotation_type = anno.type ? find_annotation_type_id(anno.type) : nil
```
afred, line 29: if `find_annotation_type_id` is returning `nil` when `anno.type.empty?`, then I don't think you need this conditional.

PR #983, opened 2025-03-18
I149 contributor bugs

laritakr:
## Summary
Refs: https://github.com/notch8/ams/issues/149

Fixes CRUD actions on a Contribution Resource within the "Credits" section of the Asset edit form. 
Removes the ability to remove members on the relationships tab, as this is NOT desired behavior for GBH, and it was breaking the ability to delete a Contribution Resource. Removing a member would not have deleted the item, it simply would have broken the relationship.

Changes can all be made in the same save, as shown in the attached video. 

<details>
<summary>Video of CRUD actions</summary>

[recorded (10).webm](https://github.com/user-attachments/assets/04640a3c-c444-456b-8417-7963dd405cb8)

</details>

## Notes

Due to how the presenter is accessing the Contribution Resources, they do not show in member id order on the show page. This is existing behavior so we did not modify it. 

## Suggested follow up work:
- Clean up Contribution Resource form, and remove basic metadata from the object, putting only the necessary metadata in the contribution resource yaml. The contribution resource as used is not a full-fledged resource with access controls, etc since it is being created in a step of the asset transaction rather than by going through the form. This means the form is never used and unnecessary. 
- Modify asset resource presenter method  `list_of_contribution_ids_to_display`. This is using authorization to create the list of ids of objects to display. As ContributionResources do not have access controls, a simpler method would be to use the member objects which are Contributions/Contribution Resources, which would also result in keeping them in sequence.


afred (approved):
👍

PR #987, opened 2025-03-20
renames scientist-softserv => notch8

ShanaLMoore:


laritakr (approved):
Good catch. I only noticed it in Gemfile!

PR #989, opened 2025-03-20
I148 fix contributor batch ingest

kirkkwang:
## Remove `affiliation` from the removed properties

151ec039266d7473051485bcac4c108ead990a1b

This commit will remove affiliation from the form's removal, which means
it's back.  I think it was erroneously added in there to begin with.

## Fix a few fixtures

47f8b0313c6b4f75a1d96d8d8214194dbb57790b

This commit was started because I wanted to write tests and found errors
in the fixtures.  Ultimately, I didn't need a spec but thought I'd
include this anyway.

## Handle contribution object differently

97f2c9d00f72ccf2b7192d8ba46b807638fc21db

Since we don't want the Contribution objects to show in the view, it
shouldn't be handled via the forms.  Instead, we want it to be handled
as if it was being added through an Asset Resource form.  There's
currently a "Credits" dropdown that now handles this through the
handle_contributors transaction.

By adding the Contribution data to the formatted_row_data before it is
sent off to the make a Hyrax::BatchIngest::BatchItem, we can inject the
information there.

## Handle destroying current ContributionResources

44faa052d3a77044219c3d55b655127951ab2158

This commit will add logic for updates to destroy the current
ContributionResources.  In an update, we want to replace what's there
while in an add we want to add to what is already there.

## Add Contribution to batch_ingest.yml

1db0f4e6fee3112273e64745001bc5ac5cf59746

This commit will add the Contribution object_class to the
batch_ingest.yml so now it passes the validation.  We don't need an id
because we are creating brand new ContributionResource objects.

### DIFF
--- a/app/forms/contribution_resource_form.rb
+++ b/app/forms/contribution_resource_form.rb
@@ -26,7 +26,6 @@ class ContributionResourceForm < Hyrax::Forms::ResourceForm(ContributionResource
   end
   remove(
     %i(
-      affiliation
       based_near
       bibliographic_citation
       creator
--- a/app/services/aapb/batch_ingest/csv_item_ingester.rb
+++ b/app/services/aapb/batch_ingest/csv_item_ingester.rb
@@ -91,6 +91,7 @@ module AAPB
             end
 
             if model_object.is_a?(AssetResource)
+              destroy_current_contributors(model_object)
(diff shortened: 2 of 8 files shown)


laritakr (approved):
good work!

ShanaLMoore:
Issue: https://github.com/notch8/ams/issues/148

PR #1017, opened 2025-04-11
minor various bug fixes

sephirothkod:
bug fix for ingesting annotations with empty fields that are set to nil. also added a spacing fix.

### DIFF
--- a/app/transactions/ams/steps/create_aapb_admin_data.rb
+++ b/app/transactions/ams/steps/create_aapb_admin_data.rb
@@ -74,7 +74,7 @@ module Ams
         return if change_set.fields["annotations"].nil?
         change_set.fields["annotations"].each do |annotation|
           next if annotation["_destroy"] == "true"
-          next if annotation.except("id").values.all? { |val| val.blank? }
+          next if annotation.except("id").values.all?(&:blank?)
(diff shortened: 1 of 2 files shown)


app/transactions/ams/steps/create_aapb_admin_data.rb
```diff
@@ -74,7 +74,7 @@ def set_annotations_attributes(admin_data, change_set)
         return if change_set.fields["annotations"].nil?
         change_set.fields["annotations"].each do |annotation|
           next if annotation["_destroy"] == "true"
-          next if annotation.except("id").values.all?(&:empty?)
+          next if annotation.except("id").values.all? { |val| val.nil? || val.empty? }
```
kirkkwang: Non-blocking but I think instead of 
```rb
next if annotation.except("id").values.all? { |val| val.nil? || val.empty? }
``` 
you can shorten it to 
```rb
next if annotation.except("id").values.all?(&:blank?)
```
since in rails we can call `#blank?` on `nil`.

kirkkwang (dismissed):
just left a non-blocking comment

app/transactions/ams/steps/create_aapb_admin_data.rb
sephirothkod: Yes of course! I used blank in my other PR, don't know why I didn't use it hear

kirkkwang (approved):

PR #1022, opened 2025-04-21
check for empty fields before assembling the pbcore xml

sephirothkod:
# Summary
When parsing the pbcore xml for export there would be a number of empty tags under the instantiations tags. Basically if the system saw an empty string it counted that as a string and would add it to the generated xml resulting in the (sometimes multiple) empty tags appearing. I rewrote how the xml is generated so that assets, all instantiations, and essence tracks all use the same method to add a field to the xml rather than every work type using a different method. this should make it easier to diagnose issues in the future.

## Ticket Number
https://github.com/notch8/ams/issues/141

# Screenshots / Video

# Expected Behavior

- go to an asset show page (asset must have instantiations with empty fields)
- add .xml to the id in the url
- see the pbcore xml with no empty tags

# Notes


sephirothkod:
https://github.com/notch8/ams/issues/141

ShanaLMoore (approved):

PR #1023, opened 2025-04-21
created new renderers for title (tesim) and date (drsim)

sephirothkod:
# Summary
This PR fixes the issue where clickable **titles** and **dates** on the assets **show page** were not linking to a proper faceted search for related objects. This PR adds new renderers for those links that adds a correct link based on the value clicked.

## Ticket Number
https://github.com/notch8/ams/issues/157

# Screenshots / Video

# Expected Behavior

- Click a title or date facet link on the asset show page
- see related objects in a standard catalog search page

# Notes


ShanaLMoore (commented):
This works but I'm wondering why faceted doesn't 🤔 ref: https://github.com/samvera/hyrax/blob/double_combo_gbh_version/app/renderers/hyrax/renderers/faceted_attribute_renderer.rb

sephirothkod:
> This works but I'm wondering why faceted doesn't 🤔 ref: https://github.com/samvera/hyrax/blob/double_combo_gbh_version/app/renderers/hyrax/renderers/faceted_attribute_renderer.rb

sorry I forgot mention that they have set up their titles to be _tesim and not _sim. so none of their titles will work with the faceted renderer unfortunately.

ShanaLMoore:
> > This works but I'm wondering why faceted doesn't 🤔 ref: https://github.com/samvera/hyrax/blob/double_combo_gbh_version/app/renderers/hyrax/renderers/faceted_attribute_renderer.rb
> 
> sorry I forgot mention that they have set up their titles to be _tesim and not _sim. so none of their titles will work with the faceted renderer unfortunately.

So an alternative solution would be for them to correct their solr index for these properties and reindex all of their data? Would that be too much trouble?

ShanaLMoore (approved):

PR #1027, opened 2025-05-12
make postgres only work

sephirothkod:
This should allow the turning off of fedora. the container can be turned off and the system should still function normally including uploads and updating records


sephirothkod:
a number of tests are failing due to af (activeFedora) not being used, yet af calls being made from tests.

sephirothkod:
eliminated af specific tests, and adapted others to use valkyrie

kirkkwang (approved):

PR #1028, opened 2025-05-13
Fix `Contribution` internal resource bug

kirkkwang:
Given a Contribution that hasn't been migrated to a Valkyrie Resource, when we look at the contributor attrs, we already have an internal resource but that needs to be deleted so it can be automatically generated otherwise we set the string like "#<Class:0x00...>" as the `internal_resource`.  The `internal_resource` should be "Contribution".

Ref:
- https://github.com/notch8/ams/issues/151


sephirothkod:
Does the persister automatically add the string "Contribution"? so if we just remove it, it should auto fill?

kirkkwang:
@sephirothkod that is correct and if we manually set it it looks like it doesn't and we're manually setting it to that string of an object and that's what makes it break it looks like

sephirothkod (approved):

PR #1030, opened 2025-05-14
changes PI format placeholder to blank space instead of a message.

sephirothkod:
for csv ingests

### DIFF
--- a/app/services/aapb/batch_ingest/csv_item_ingester.rb
+++ b/app/services/aapb/batch_ingest/csv_item_ingester.rb
@@ -78,7 +78,7 @@ module AAPB
           attributes["admin_set_id"] = @batch_item.batch.admin_set_id
 
           unless %w[Asset DigitalInstantiation EssenceTrack].include?(node.object_class)
-            attributes["format"] = attributes["format"].present? ? attributes["format"] : 'Instantiation format not provided'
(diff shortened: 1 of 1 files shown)


app/services/aapb/batch_ingest/csv_item_ingester.rb
```diff
@@ -78,7 +78,7 @@ def ingest_object_at(node, with_data, with_parent = false)
           attributes["admin_set_id"] = @batch_item.batch.admin_set_id
 
           unless %w[Asset DigitalInstantiation EssenceTrack].include?(node.object_class)
-            attributes["format"] = attributes["format"].present? ? attributes["format"] : 'Instantiation format not provided'
+            attributes["format"] = attributes["format"].present? ? attributes["format"] : '   '
```
aprilrieger, line 81: Non blocking question, did you mean to leave empty space here? `' '` Or does it just look weird like this in GitHub?

aprilrieger (approved):

app/services/aapb/batch_ingest/csv_item_ingester.rb
sephirothkod: We literally are putting a space there to be an "invisible" placeholder. So it is actually meant to be there.

PR #1039, opened 2025-05-23
Adds total and numbers (not bullets) for Push details page

afred:


afred:
specs originally failed for reasons unknown, but re-running them succeeded 🤷 . Here's a visual of the change

image

ekemeyer (approved):
helpful update - catalogers already on board with it too

PR #1055, opened 2025-06-27
Add rundown

### LINKED ISSUE #1052: additional description type in AMS2
# Details 
Request: Can we have an additional type of pbcoreDescription in AMS2?

Context:  For the CD transfer project, we have tracklists from the CDs and program rundowns.  These are time-based indexes of the program content.  The consensus in the cataloging team is that ideally these belong in pbcoreDescription, but it should not be in the main description.  We need a pbcoreDescription of a different type. 

Here are is the official PBCore controlled vocabulary for @descriptionType:
https://pbcore.org/pbcore-controlled-vocabularies/descriptiontype-vocabulary/

We think the right term for these tracklists would be "Rundown".  However, "Rundown" is not one of the currently available options.  The current options are in this file:
https://github.com/WGBH-MLA/ams/blob/develop/config/authorities/description_types.yml

If we add "Rundown" to that list, will that allow us to use that attribute value?

Submitted by: Owen, Lauren, and CD transfer interns
CC in communications: rebecca_fraimow@wgbh.org
Priority: Medium (within this month)
URL: https://github.com/WGBH-MLA/ams/blob/develop/config/authorities/description_types.yml
Slack message thread:

ekemeyer:
Closes #1052 

Needs to be available:

- [x]  Manually in individual records
- [x] Batch Asset Ingest
- [x] Batch Update Attribute Ingest
- [x] Batch Add Attribute Ingest

Needs to be able to:

- [x] Push to AAPB successfully


ekemeyer:
Erica local tests:
No errors when passing `SETTINGS__BULKRAX__ENABLED=true`

Following errors when passing `SETTINGS__BULKRAX__ENABLED=false`
```
rspec ./spec/models/bulkrax/pbcore_xml_entry_spec.rb:91 # Bulkrax::PbcoreXmlEntry#build with raw_metadata builds entry
rspec ./spec/models/bulkrax/pbcore_xml_entry_spec.rb:86 # Bulkrax::PbcoreXmlEntry#build with raw_metadata succeeds
```

afred (approved):
👍

PR #1056, opened 2025-07-01
Export sonyci

### LINKED ISSUE #1057: Add sonyci to asset exports
**Is your feature request related to a problem? Please describe.**
Clunky process being used to pull down AMS2 data and parse it to identify sony ci IDs by collection

**Describe the solution you'd like**
Simply provide the sony ci in the csv export for assets, this way the search/filtering can be prepared in UI

**Describe alternatives you've considered**
Pulling down JSON based on curl queries and then parsing it for relevant information

ekemeyer:
Adds sonyci as an exported field for assets in a csv.

Closed #1057

### DIFF
--- a/app/controllers/pushes_controller.rb
+++ b/app/controllers/pushes_controller.rb
@@ -12,7 +12,7 @@ class PushesController < ApplicationController
   end
 
   def create
-    @push = Push.create(user_id: current_user.id, status: 'pending', pushed_id_csv: pushed_id_csv_from_id_field)
+    @push = Push.create(user_id: current_user.id, status: 'pending')
     if @push.valid?
       SavePushJob.perform_later(push: @push, pushed_id_csv: pushed_id_csv_from_id_field)
       PushToAAPBJob.perform_later(id: @push.id, user: current_user)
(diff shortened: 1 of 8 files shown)


app/controllers/pushes_controller.rb
```diff
@@ -12,7 +12,7 @@ def show
   end
 
   def create
-    @push = Push.create(user_id: current_user.id, status: 'pending')
+    @push = Push.create(user_id: current_user.id, status: 'pending', pushed_id_csv: pushed_id_csv_from_id_field)
```
afred: Here I think we need to leave the `pushed_id_csv: pushed_id_csv_from_id_field` to line `#17` below. This is probably an artifact of the merge.

docker-compose.yml
```diff
@@ -239,6 +239,8 @@ services:
         condition: service_completed_successfully
     expose:
       - 3000
+    ports:
+      - "3000:3000"
```
afred: Leave this one off. This is confusing, I know, but it conflicts with the Notch8 way of port forwarding in their dev environments.

What I tend to do when I'm developing locally is add these port forwarding lines to my working copy of `docker-compose.yml` but then just try to remember not to commit it them.

afred (changes requested):

docker-compose.yml
ekemeyer: Ah yea, I definitely forgot to not commit this

afred (approved):
👍

PR #1060, opened 2025-07-03
Remove annotation type requirement

### LINKED ISSUE #1026: Allow PBcore asset annotation without a type for PBCore Zipped
It is valid pbcore to have an asset annotation without a specific type. Currently this is not possible when ingesting pbcore via PBCore Zipped ingester. There are many records waiting to go into AMS2 that do not have a type.

ekemeyer:
PBCore XML files remaining to be migrated into AMS2 have no annotationTypes, which is valid PBCore, but we have a check/requirement for it. Removing this to allow migration of PBCore xml.

closes #1026


ekemeyer:
Tests are all passing locally with Bulkrax enabled.

With Bulkrax disabled the instantiations are not all getting added in my local environment. However, they all get ingested on demo with this branch and everything looks good.

afred (approved):
👍 looks good!

ekemeyer:
Working on demo, however it should be noted you are not able to manually add annotations without types at this point.

PR #1077, opened 2025-07-28
Avoid error Asset  description is missing

afred:


ekemeyer (approved):
Learned about `&` today - looks good! Thanks Drew.

PR #1109, opened 2025-11-18
Updates for upgrade to Amazon Linux 2023

maxkadel:
- Use bitnamilegacy image for Redis
- Add more waits to the ssh-keyscan to wait for DNS to resolve in pods


afred (dismissed):
👍 these look good, although I'm wondering if we still need the alpha,bravo,charlie,delta deployments anymore. These were for migration to widen the pipe into AMS (it appears alpha and bravo were for prod, and charlie and delta were for demo), but I think we are done with that whole process now and we can tighten up the k8s cluster by removing those deployments and dependent resources.

afred (approved):
👍

PR #1110, opened 2025-11-18
Change DB_HOST from 'postgres' to 'db'

aprilrieger:
Change DB_HOST from 'postgres' to 'db' to resolve to the DB docker container during tests.


maxkadel (changes requested):
I figured it out - the "db" is for the mariadb backing of Fedora, not the Rails application's database. The postgres container was using the `latest` tag, which switched from v17 to v18 earlier this year, and had breaking changes.

PR #1114, opened 2025-12-11
fixes failure and success silent delivery failure

sephirothkod:


app/jobs/export_records_job.rb
```diff
@@ -21,10 +21,9 @@ class ExportRecordsJob < ApplicationJob
   #   #named_arguments (see ApplicationJob#named_arguments).
   def perform(export_type:, user:, search_params: {})
     delivery.deliver
+    notification.send_success
```
afred, line 24: Does this mean that in the case of a failure on `delivery.deliver` we'll get both a call to `notification.send_success` first followed by a call to `failure_notification.send_failure` in the `rescue` clause above?

sephirothkod: no it shouldn't send two notifications as long as an error is thrown in the `delivery.deliver` call; it should stop immediately when the error is thrown and not continue with the perform method. This is the same set up that we have had in the AAPB push job (below) since the summer, we just didn't realize at the time that this job used the same flawed `after_perform { notification.send_success }` (which would send the two notifications) so I fixed it here too.

PR #1123, opened 2026-02-03
Removes Rancher from required providers from TF template

### LINKED ISSUE #1122: Remove Rancher from terraform template
### Because
We need to uninstall Rancher agents from our K8s cluster, and the terraform temlplate is likely how they got installed.

If we use the Terraform tempalates for provisioning in the future (tbd) we won't want the Rancher dependencies installed.

### Done when
Ranger dependencies are removed from Terraform template

afred:
Closes #1122


afred:
covered by #1125

PR #1143, opened 2026-03-06
Adds API endpoint for showing raw indexed search doc

afred:


foo4thought (changes requested):
this looks good!  does it accept any other parameters like fl (field list) as does the api endpoint on AAPB?

foo4thought (dismissed):
disregard my question in the other comment

foglabs (dismissed):

PR #1155, opened 2026-04-29
Contributor updates

ekemeyer:
Changes based on [document](https://docs.google.com/document/d/1RSpxsEJJco1tgliJ__Pm-zuSL53N98TdqmHKcXdaY4A/edit?tab=t.0) provided by Owen. This PR covers the following:

1. Support one additional PBCore attribute in the contributor element: `annotation`
2. Add 2 new items to the role controlled vocabulary:
`Appearing`: an on-air contributor (like cast)
`Crew`: a contributor the the production
3. Fixes existing bug that ignored `affiliation` and `portrayal` inputs. These should now map into pbcore/xml.

Still to do:

- [ ] Test push of these types of records to AAPB

### DIFF
--- a/app/services/aapb/batch_ingest/pbcore_xml_mapper.rb
+++ b/app/services/aapb/batch_ingest/pbcore_xml_mapper.rb
@@ -161,11 +161,12 @@ module AAPB
 
       def person_attributes(person, role)
         {
-          contributor: (person.value if person),
+          contributor:      (person.value if person),
           contributor_role: (role.value if role),
           # pbcorecontributor ONLY
-          affiliation: (person.affiliation if defined? person.affiliation),
-          portrayal: (role.portrayal if role && defined? role.portrayal),
+          affiliation:      (person.respond_to?(:affiliation) ? person.affiliation : nil),
+          portrayal:        (role.respond_to?(:portrayal) ? role.portrayal : nil if role),
+          annotation:       (person.respond_to?(:annotation) ? person.annotation : nil)
         }
 
       end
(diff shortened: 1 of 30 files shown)


app/services/aapb/batch_ingest/pbcore_xml_mapper.rb
```diff
@@ -161,11 +161,12 @@ def people_attributes(people)
...
       def person_attributes(person, role)
         {
-          contributor: (person.value if person),
+          contributor:      (person.value if person),
           contributor_role: (role.value if role),
           # pbcorecontributor ONLY
-          affiliation: (person.affiliation if defined? person.affiliation),
-          portrayal: (role.portrayal if role && defined? role.portrayal),
+          affiliation:      (person.respond_to?(:affiliation) ? person.affiliation : nil),
+          portrayal:        (role.respond_to?(:portrayal) ? role.portrayal : nil if role),
```
afred: is that `if role` needed in there? Looks like a relic from an edit.

spec/models/contribution_spec.rb
```diff
@@ -42,12 +42,22 @@
         expect(contribution.portrayal.include?("Test portrayal")).to be true
       end
     end
+
     context "affiliation" do
       it "has affiliation" do
-        contribution.portrayal = "Test affiliation"
+        contribution.affiliation = "Test affiliation"
```
afred, line 48: Looks like this test was testing the wrong thing before, yeah?

app/services/aapb/batch_ingest/pbcore_xml_mapper.rb
ekemeyer: good catch, removing

ekemeyer:
Latest changes to get test passing - prevent Rails from trying to execute real delayed scheduling while still verifying that the job makes an attempt to reschedule itself.

afred (approved):
👍 reviewed the commit and leaning on my previous approval for the rest :) Your code is looking great!

PR #1162, opened 2026-05-19
Update NGINX annotations

### LINKED ISSUE #1141: Increase NGINX Ingress Limit
Raananah and Miranda ran into a `413 Request Entity Too Large` nginx error when trying to upload a 1.3MB zipped file for a digital instantiation ingest.

This should be raised to 2MB - set probably in a yaml file in cluster somewhere?

ekemeyer:
- Updated `proxy-body-size` to `client-max-body-size` since it did not seem like proxy body size was being rendered when I increased it from 1mb to 2mb. We are using `NGINX Ingress Controller Version=5.3.3`, which (according to Claude) uses `client-max-body-size` annotation, as opposed to `ingress-nginx`, which uses `proxy-body-size`. I'm assuming when we had `proxy-body-size=1m` it was not doing anything and the default 1m was in play, so it wasn't noticed.
- Increased limit from 1mb to 2mb
- Added check and notice to user that appears before submitting an ingest that is larger than 2mb
- Added read/send timeouts to 2 minutes. 

This closes #1141 and avoids `entity too large` errors on ingests between 1-2MB, which catalogers have been experiencing.

### DIFF
--- a/app/views/hyrax/batch_ingest/batches/new.html.erb
+++ b/app/views/hyrax/batch_ingest/batches/new.html.erb
@@ -10,3 +10,19 @@
   <%= f.input :admin_set_id, collection: @admin_sets, required: true %>
   <%= f.button :submit,  value: 'Save' %>
 <% end %>
+
+<script>
+  document.addEventListener('DOMContentLoaded', function() {
+    const form = document.querySelector('form');
+    const fileInput = document.querySelector('input[type="file"]');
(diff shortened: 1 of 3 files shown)


ekemeyer:
Tested demo with this update, and it does accept larger ingests. Will need to communicate to users that there is a longer wait period that they should expect if they have larger files.

afred (approved):
👍

PR #1182, opened 2026-07-10
Admin data fix

ekemeyer:
Catalogers were ingesting digital instantiation pbcore xml files via the zipped digital instantiation ingester, alongside a manifest.csv file containing additional metadata for each xml (generation, holding_organization, md5, preservation lto, preservation disk). The preservation data (`DigitalInstantiation.aapb_preservation_lto` and `DigitalInstantiation.aapb_preservation_disk`) was getting dropped, in addition to not having an annotationType to be assigned to.

This fix maps the dropped preservation data to annotation fields in the digital instantiation. It also updates the instantiation reset ingester to map the same data appropriately when it is coming in via that route, and updates "md5" as a valid source so it is more clear in PBCore what the ID is and where it came from.

- [x]  Confirm annotationType names are OK with Rebecca before deploying

### DIFF
--- a/app/services/aapb/batch_ingest/pbcore_xml_mapper.rb
+++ b/app/services/aapb/batch_ingest/pbcore_xml_mapper.rb
@@ -205,7 +205,7 @@ module AAPB
           attrs[:channel_configuration]           = pbcore.channel_configuration&.value
           attrs[:alternative_modes]               = pbcore.alternative_modes&.value
 
-          md5_ids, other_ids = pbcore.identifiers.partition { |identifier| identifier.source.to_s.downcase.strip == "md5" }
+          md5_ids, other_ids = pbcore.identifiers.partition { |identifier| identifier.source.to_s.downcase.strip =~ /md5/ }
           attrs[:md5] = md5_ids.first.value if md5_ids.present?
           attrs[:local_instantiation_identifier] = other_ids.select { |identifier| identifier.source.to_s.downcase.strip != "ams" }.map(&:value)
 
--- a/app/services/aapb/batch_ingest/pbcore_xml_mapper.rb
+++ b/app/services/aapb/batch_ingest/pbcore_xml_mapper.rb
@@ -201,14 +201,22 @@ module AAPB
           # TODO: change left side to rights_summaries (because multiple: true), or to rights (to agree with PBCore gem's Instantiation#rights)
           attrs[:rights_summary]                  = pbcore.rights.map(&:summary).map(&:value)
           attrs[:rights_link]                     = pbcore.rights.map(&:link).map(&:value)
-          attrs[:local_instantiation_identifier]  = pbcore.identifiers.select { |identifier| identifier.source.to_s.downcase.strip != "ams" }.map(&:value)
           attrs[:tracks]                          = pbcore.tracks&.value
           attrs[:channel_configuration]           = pbcore.channel_configuration&.value
           attrs[:alternative_modes]               = pbcore.alternative_modes&.value
 
-          orgs, annotations = pbcore.annotations.partition { |anno| anno.type && anno.type.downcase == 'organization' }
-          attrs[:holding_organization] = orgs.first.value if orgs.present?
-          attrs[:annotation] = annotations.map(&:value)
+          md5_ids, other_ids = pbcore.identifiers.partition { |identifier| identifier.source.to_s.downcase.strip == "md5" }
+          attrs[:md5] = md5_ids.first.value if md5_ids.present?
+          attrs[:local_instantiation_identifier] = other_ids.select { |identifier| identifier.source.to_s.downcase.strip != "ams" }.map(&:value)
+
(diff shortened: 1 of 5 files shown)


app/services/aapb/batch_ingest/pbcore_xml_mapper.rb
```diff
@@ -201,14 +201,22 @@ def instantiation_attributes
...
           attrs[:rights_link]                     = pbcore.rights.map(&:link).map(&:value)
-          attrs[:local_instantiation_identifier]  = pbcore.identifiers.select { |identifier| identifier.source.to_s.downcase.strip != "ams" }.map(&:value)
           attrs[:tracks]                          = pbcore.tracks&.value
           attrs[:channel_configuration]           = pbcore.channel_configuration&.value
           attrs[:alternative_modes]               = pbcore.alternative_modes&.value
 
-          orgs, annotations = pbcore.annotations.partition { |anno| anno.type && anno.type.downcase == 'organization' }
-          attrs[:holding_organization] = orgs.first.value if orgs.present?
-          attrs[:annotation] = annotations.map(&:value)
+          md5_ids, other_ids = pbcore.identifiers.partition { |identifier| identifier.source.to_s.downcase.strip == "md5" }
```
afred: On this line, it appears the identifier source is expected to always be "md5" or "MD5"? Do you know if there are any other possibilities? I only ask because sometimes I see references to "md5 checksum" and not sure if identifier source is specified that way sometimes.

If you're sure it's always md5, then 👍 . If not, you could use a regex here to catch _anything_ with "md5" in the string anywhere.

app/transactions/ams/steps/create_aapb_admin_data.rb
```diff
@@ -204,8 +204,13 @@ def find_or_create_instantiation_admin_data(change_set)
 
       def set_instantiation_admin_data_attributes(change_set)
         instantiation_admin_data_attributes.each do |k|
-          change_set.instantiation_admin_data.send("#{k}=", change_set.fields[k].to_s)
+          value = change_set.fields[k]
+          value = change_set.fields[k.to_s] if value.nil?
```
afred, line 208: I take it this is because keys to `change_set.fields` are sometimes keys and sometimes symbols?

app/services/aapb/batch_ingest/zipped_pbcore_reader.rb
```diff
@@ -25,7 +25,7 @@ def unzipped_xml_file_paths
 
         def unzipped_file_paths
           @unzipped_file_paths  ||= begin
-            files = zip_file.glob('**/*.xml') + zip_file.glob('**/*.pbcore')
+            files = zip_file.glob('**/*.xml') + zip_file.glob('**/*.pbcore') + zip_file.glob('**/*.csv')
```
afred: Why is the `.csv` blob being added here?

app/services/aapb/batch_ingest/pbcore_xml_mapper.rb
ekemeyer: for our own internal/mellon-based instantiation work it will always be "md5" 

see https://github.com/WGBH-MLA/ams/blob/d331bb3db371f2b00a43db26059b770cdbc8ab47/app/models/ams/pbcore_xml_export_extension.rb#L213

but you're right that eventually or from legacy projects it could look different. I'll add the check for anything with md5.

app/services/aapb/batch_ingest/zipped_pbcore_reader.rb
ekemeyer: great question! It was my initial, hopeful, attempt to resolve the error. Was not sure if the .csv was getting read on ingest, but it is here https://github.com/WGBH-MLA/ams/blob/d331bb3db371f2b00a43db26059b770cdbc8ab47/app/services/aapb/batch_ingest/zipped_pbcore_digital_instantiation_reader.rb#L21:

removing my change from next push

afred: I followed the code, and pretty sure that only `.pbcore` and `.xml` are getting filtered through in the end, but i do think taking the `.csv` out of that group of files will be less confusing in the future.

app/transactions/ams/steps/create_aapb_admin_data.rb
afred: if so.. this is fine, i'm just thinking that I would expect the keys of `change_set.fields` to either be consistent, or to be a hash `with_indifferent_access` where you can use either. If it's not consistent, nor having indifferent access, then I'm guessing that the Hyrax code is just honoring whatever key type was used to define the fields, in which case _we_ might be the inconsistent ones somewhere in our config. Not crucial, but something to keep an eye out for and maybe make consistent (preferably symbols, imo) if we find an opportunity.

ekemeyer: I am not entirely certain where the inconsistency started, but this is the key change that fixes the bug with admin instantiation data not getting saved.

afred (approved):
👍

